PR #17, opened 2020-12-07
Kubernetes Mock API

einari:
This PR introduces a mock for "just enough" of the Kubernetes API. Meaning, for now the 2 routes we need (namespaces + pods). On the returned data, there is also just enough what we need to be able to get the initial stories completed.

# How to run it

The environment is now updated to redirect '/api/v1' to the K8sMock. Restart the environment for this if already running.
The K8sMock needs to be running to be able to use this, so `yarn start` in the K8sMock folder.
With this, one can now go to the GraphQL endpoint and perform the Applications query and get data:

image

# Swagger

The K8sMock service exposes a Swagger endpoint at `http://localhost:3001/api/swagger/`. Navigate to it and try out:
image

# Mock Data

The data being returned sits in `pods.json` in the `pods` folder next to the `PodsController` in the K8sMock project and in `namespaces` in the `namespaces` folder next to the `NamespacesController` in the K8sMock project.

# Next Steps

With the PodsController in the K8sMock already producing data, the natural next step is to use this. 
I suggest creating a new object that combines the Application and its Microservices as an array on it. Something like 'ApplicationsForListing' or similar and then a query for this purpose. In the future this is what we would project to as well.


Source/Applications/Backend/applications/ApplicationQueries.ts
```diff
@@ -6,6 +6,11 @@ import { ILogger } from '@shared/backend/logging';
 import { injectable } from 'tsyringe';
 import { Application, ApplicationModel } from './Application';
 
+import { NamespaceList } from '@shared/backend/k8s/NamespaceList';
```
pavsaund, line 9: I get a squiggly in vscode under this import with the following message:
>Output file '/Users/pavneet/code/Dolittle/Studio/Source/Shared/Backend/dist/k8s/NamespaceList.d.ts' has not been built from source file >'/Users/pavneet/code/Dolittle/Studio/Source/Shared/Backend/k8s/NamespaceList.ts'.ts(6305)

It builds just fine, but I suspect we need to export the k8s shared modules from the backend?

Source/K8sMock/namespaces/NamespacesController.ts
```diff
@@ -0,0 +1,17 @@
...
+
+@Route('api/v1/namespaces')
+@Tags('core_v1')
+export class NamespacesController extends Controller {
+
+    @Get()
+    async getNamespaces(): Promise<PodList> {
+        return namespaces;
+    }
+}
```
pavsaund, line 17: I presume `PodList` references here, should be `NamespaceList`?

pavsaund (changes requested):
This seems straight-forward enough. Could you have a look at my comments and see if they make sense?

Source/Applications/Backend/applications/ApplicationQueries.ts
einari: I need to do some investigations on this. I saw this yesterday locally as well and did a `yarn tsc -b` in the Shared/Backend folder and it went away. My guess is that something is not configured correctly. 
I'll dig a bit.

Source/K8sMock/namespaces/NamespacesController.ts
einari: :) Indeed.. Thnx.. Will fix..

Source/Applications/Backend/applications/ApplicationQueries.ts
pavsaund: oh! `yarn tsc -b` fixed it for me as well. It also fixed the `withProjectionFor` squiggles I've been getting. So is there something with the typescript projects that isn't updating the built shared projects?

einari:
The problem with the squigglies are related to this: `disableSourceOfProjectReferenceRedirect`- I see that this was added due to this: https://github.com/microsoft/TypeScript/issues/34677.

Source/Applications/Backend/applications/ApplicationQueries.ts
pavsaund: Solving this outside this PR

pavsaund (approved):

PR #36, opened 2020-12-15
List applications and microservices in the navigation pane

pavsaund:
This PR will introduce the ability to list out Microservices for each application in the Navigation pane. The structure is not yet clickable.
- ApplicationForListing and MicroserviceForListing: Introduced specific models and queries for listing objects that will contain just enough information for that purpose.
- K8sClient: A client wrapper that talks to the k8smock API for namespaces and pods. This will need to be evolved
- disabled action bar component directly for now, as we don't want to be able to create a microservice


Source/Applications/Web/ApplicationForListingModel.ts
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Dolittle. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+import { Guid } from '@dolittle/rudiments';
+import { MicroserviceForListing } from './MicroserviceForListing';
+
+export type ApplicationForListingModel = {
```
jakhog, line 6: Just for consistency - should this be called `ApplicationForListing`? The other one is called `MicroserviceForListing` and the `...Model` suffix is usually used on the `...ViewModel` classes.

Source/Applications/Backend/listing/ApplicationForListing.ts
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Dolittle. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+import { Field, ObjectType } from 'type-graphql';
+import { guid } from '@shared/backend/data';
+import { prop, getModelForClass } from '@typegoose/typegoose';
+import { Guid } from '@dolittle/rudiments';
+import { MicroserviceForListing } from './MicroserviceForListing';
```
jakhog, line 8: We don't need all of these:
```suggestion
import { Field, ObjectType } from 'type-graphql';
import { prop, getModelForClass } from '@typegoose/typegoose';
import { MicroserviceForListing } from './MicroserviceForListing';
```

jakhog (changes requested):
Just two tiny things

Source/Applications/Web/ApplicationForListingModel.ts
pavsaund: Good catch - I missed that, I cannot remember my reasoning for suffixing with Model, but I will remove it.

jakhog (approved):

PR #38, opened 2020-12-17
Expand K8sMock and hook up K8sClient to fetch applications and microservices

jakhog:
We have
- Removed some old unused code in the K8sMock (events++)
- Added a few more controllers so that we can get Namespaces, Deployments, Pods, ConfigMaps from the K8sMock
- Added more data, and aligned it with what it looks like in production
- Replaced the custom KubeClient with the official one, so that we don't have to manually map all the types
- Enabled configuration of the Backends based on environment variables (we can make it more fancy later)


Source/K8sMock/configmaps/ConfigmapsController.ts
```diff
@@ -0,0 +1,322 @@
...
+
+@Route('api/v1/namespaces')
+@Tags('core_v1')
+export class ConfigmapsController extends Controller {
+    private readonly _configmaps: V1ConfigMap[] = [];
+
+    constructor() {
+        super();
+
+        {
```
joelhoisko, line 17: Some quirky identation happening here

Source/Shared/Backend/k8s/MicroserviceResources.ts
```diff
@@ -0,0 +1,67 @@
...
+            undefined,
+            undefined,
+            {
+                headers: {
+                    'User-ID': ctx.userId,
+                    'Tenant-ID': ctx.tenantId,
+                }
+            });
+        return body.items;
+    };
```
joelhoisko, line 44: We'd probably want to check the returned response in case of some failure and have different logging for them/errors to throw, something along the lines of having different errors if they come from the impersonator or kubernetes.

joelhoisko (commented):
Minor comment regarding error handling.

PR #54, opened 2021-02-09
Studio v3 friendly

freshteapot:
Making Studio cluster V3 friendly.
Important part was gettings applications via the new tenant-id path.

A little bit of boilerplate replacing

# Todo
- [ ] Do we need to update vanir?
- [ ] The read_models db name is not the same as in the cluster
- [ ] The event_store name is not the same as in the cluster


woksin (approved):
This seems fine to me. I don't have the insight to know whether there are more files that needs to be changed, but at least the files that has been changed here seems to be done correctly

PR #55, opened 2021-02-15
Connecting golang backup service with Studio graphql

freshteapot:
**To run locally with the go server and to talk to azure, you will need admin access to cluster three.**

Linked with https://github.com/dolittle-platform/Operations/pull/13

- server is in https://github.com/dolittle-platform/Operations/pull/13.

# Todo
- [x] Fix all frontends to actually run
- [x] Investigate why the backend complains about controllers, **it is complaining because there are no files with Controller Suffix**.
- [x] Get an overview of the abstraction going on from "web" -> "backend" -> data source.

# How to
- setup yarn
```sh
yarn
```
- start mongo
```sh
cd .dolittle
docker-compose up -d studio-mongo
```

- start backup
```sh
cd ./Source/Data/Backend
yarn start:dev
```

- start web
```sh
cd ./Source/Data/Web
yarn start:dev
```

Linked to https://app.asana.com/0/1199900680512261/1199900680512276/f


goeran:
@freshteapot I am asking to learn. I noticed that `./dolittle/data/resources.json` is a folder. Is that intentional?

freshteapot:
I haven't tried running via docker.

freshteapot:
@goeran both you and @woksin have made reference to this folder.
I do not have it.
Which makes me think it must be a side effect / feature of docker-compose.

Source/Data/Backend/package.json
```diff
@@ -30,8 +30,9 @@
         ],
         "ext": "ts"
     },
-    "dependencies": {
-        "@shared/backend": "1.0.0",
-        "@shared/specs": "1.0.0"
+    "devDependencies": {
+        "@apollo/client": "^3.3.11",
```
einari, line 34: Is this dependency needed?

einari (commented):
Looking good.
Just wondering about the devDependency to the client library in the backend.

einari:
As for the checkbox you have for the controllers and why we get that error. Tsoa returns this when it's trying to generate it's config and Swagger from the code and there are no controllers.

Could be that there is way to tell it not to complain or simply take it out of the pipeline till one has APIs to expose.

Source/Data/Backend/package.json
freshteapot: I will double check

freshteapot: Removed and at least locally seems happy.

freshteapot:
> As for the checkbox you have for the controllers and why we get that error. Tsoa returns this when it's trying to generate it's config and Swagger from the code and there are no controllers.
> 
> Could be that there is way to tell it not to complain or simply take it out of the pipeline till one has APIs to expose.

I am going to let it be for now, as it is only in development and until we have more shared knowledge I or someone else might re-bootstrap it in an odd way :).

freshteapot:
@einari sorry for not following up on this PR.
I have fixed the things you bubbled up, can I get a tick?

einari (approved):

PR #56, opened 2021-02-17
Add resources configuration to Data under .dolittle so that docker-compose works

woksin:
The Data microservice was missing the resources configuration  so the docker-compose would not run.

So either the resources.json was missing or the docker-compose was not supposed to spin up the runtime for the Data microservice yet.


woksin:
@einari you probably know what's right here :)

einari (approved):
Looks exactly like the one I have on my machine. And the reason it wasn't in the repo is that "Data" is in .gitignore :)

PR #59, opened 2021-02-23
Fix build pipline ACR login failure

jakhog:
Using the official azure/docker-login action that better handles crazy characters in the password.


freshteapot (approved):
Approved!

PR #64, opened 2021-03-24
Updating configmaps to use __

freshteapot:
- Only fixing "Data" as I can't quite get the lay of the land, yet.
- Development got out of sync, my fault as I am learning to keep the configmap in here.
- Moving x-secret to be configured via ENV.


freshteapot:
@einari I will merge this in, as its currently in the cluster, have left dev-applications, which is broken as I can't quite figure out what is correct. IE prod works but dev does not. I think we need to update them all to "__", but I have a small thought that maybe you were fixing this in the code.

PR #69, opened 2021-04-19
Initial scaffold of Business Moments Adapters

einari:


goeran:
@einari when I try to query for all entities I get this error message,

```
{
  "errors": [
    {
      "message": "Cannot inject the dependency \"_mongoDatabase\" at position #0 of \"EntityQueries\" constructor. Reason:\n    Missing resource configurations for '445f8ea8-1a6f-40d7-b2fc-796dba92dc44'",
      "locations": [
        {
          "line": 3,
          "column": 5
        }
      ],
      "path": [
        "entities",
        "allEntities"
      ],
      "extensions": {
        "code": "INTERNAL_SERVER_ERROR",
        "exception": {
          "stacktrace": [
            "Error: Cannot inject the dependency \"_mongoDatabase\" at position #0 of \"EntityQueries\" constructor. Reason:",
            "    Missing resource configurations for '445f8ea8-1a6f-40d7-b2fc-796dba92dc44'",
            "    at /Users/gh/code/dolittle/Studio/node_modules/tsyringe/dist/cjs/dependency-container.js:290:23",
            "    at Array.map (<anonymous>)",
            "    at InternalDependencyContainer.construct (/Users/gh/code/dolittle/Studio/node_modules/tsyringe/dist/cjs/dependency-container.js:266:34)",
            "    at InternalDependencyContainer.resolve (/Users/gh/code/dolittle/Studio/node_modules/tsyringe/dist/cjs/dependency-container.js:105:33)",
            "    at Object.get (/Users/gh/code/dolittle/Studio/node_modules/@dolittle/vanir-backend/graphql/GraphQLSchema.ts:39:34)",
            "    at IOCContainer.getInstance (/Users/gh/code/dolittle/Studio/node_modules/type-graphql/dist/utils/container.js:39:26)",
            "    at /Users/gh/code/dolittle/Studio/node_modules/type-graphql/dist/resolvers/create.js:15:51",
            "    at field.resolve (/Users/gh/code/dolittle/Studio/node_modules/apollo-server-core/src/utils/schemaInstrumentation.ts:103:18)",
            "    at resolveField (/Users/gh/code/dolittle/Studio/node_modules/graphql/execution/execute.js:466:18)",
            "    at executeFields (/Users/gh/code/dolittle/Studio/node_modules/graphql/execution/execute.js:294:18)"
          ]
        }
      }
    }
  ],
  "data": {
    "
```

My query looks like this. And for the record, I don't think it's the query that's causing this error. It has to be my setup?

```
query {
  entities {
    allEntities {
      _id
    }
  }
}
```

einari:
That sounds like a missing resources file inside `.dolittle` folder of the Backend. Weird. Let me double check that all files are committed later.

goeran:
@einari Thanks! please do :)

einari:
Or... Hang on..  I think I know why..  The resources are configured with a different Tenant ID and we've optimized the flow to be that you go through the "ingress" in the Docker Compose - meaning that you navigate to `http://localhost:9000/_/businessmomentsadapter/graphql` rather than directly to the service. 

If you want to navigate directly to the service, I suggest installing the modheader extension to your browser and set the tenant-id header to: 508c1745-5f2a-4b4c-b7a5-2fbb1484346d. You can read more about the headers here:https://github.com/dolittle-entropy/vanir/blob/main/Documentation/microservice.md#http-headers

Source/BusinessMomentsAdapters/Backend/domain/adapters/Adapter.ts
```diff
@@ -0,0 +1,39 @@
...
+import { EventSourceId } from '@dolittle/sdk.events';
+import { Guid } from '@dolittle/rudiments';
+import { ConnectorAddedToAdapter } from '../../events/adapters';
+import { BrokenRule } from '@dolittle/vanir-backend';
+
+export const ConnectorWithNameAlreadyExists =
+    BrokenRule.create('ConnectorWithNameAlreadyExists', 'Connector with name "{name}" already exists on adapter "{adapter}"');
+
+@aggregateRoot('08dfe483-0b6a-40e1-95b7-1b3c7ccaf187')
+export class Adapter extends AggregateRoot {
```
goeran, line 14: @einari Nice example of an AggregateRoot!

goeran:
@einari I think your last comment should be added as docs. It's very valuable, and it make sense to go through the proxy.

goeran:
Screenshot 2021-04-23 at 12 10 35

@einari I get 502 Bad Gateway when trying to use the "ingress". I'm running the back-end on port 3006. I know because `http://localhost:3006/_/businessmomentsadapters/graphql` is working as expected.

Source/BusinessMomentsAdapters/Backend/read/entities/EntityQueries.ts
```diff
@@ -0,0 +1,21 @@
...
+import { Entity } from './Entity';
+import { IMongoDatabase, graphRoot } from '@dolittle/vanir-backend';
+import { injectable } from 'tsyringe';
+
+@Resolver()
+@graphRoot('entities')
+@injectable()
+export class EntityQueries {
+
+    constructor(private readonly _mongoDatabase: IMongoDatabase) { }
```
goeran, line 14: Vanir is setting up the mongodatabase object instance for me. While doing so it is also magically capturing tenante-id from the http request header. This is cool magic, but I think the error messages I get when tenant-id header is not explaining directly what's wrong.

Here is the error message when I query all entities via Graphql:

```
{
  "errors": [
    {
      "message": "Cannot inject the dependency \"_mongoDatabase\" at position #0 of \"EntityQueries\" constructor. Reason:\n    Missing resource configurations for '445f8ea8-1a6f-40d7-b2fc-796dba92dc44'",
      "locations": [
        {
          "line": 3,
          "column": 5
        }
      ],
      "path": [
        "entities",
        "allEntities"
      ],
      "extensions": {
        "code": "INTERNAL_SERVER_ERROR",
        "exception": {
          "stacktrace": [
            "Error: Cannot inject the dependency \"_mongoDatabase\" at position #0 of \"EntityQueries\" constructor. Reason:",
            "    Missing resource configurations for '445f8ea8-1a6f-40d7-b2fc-796dba92dc44'",
            "    at /Users/gh/code/dolittle/Studio/node_modules/tsyringe/dist/cjs/dependency-container.js:290:23",
            "    at Array.map (<anonymous>)",
            "    at InternalDependencyContainer.construct (/Users/gh/code/dolittle/Studio/node_modules/tsyringe/dist/cjs/dependency-container.js:266:34)",
            "    at InternalDependencyContainer.resolve (/Users/gh/code/dolittle/Studio/node_modules/tsyringe/dist/cjs/dependency-container.js:105:33)",
            "    at Object.get (/Users/gh/code/dolittle/Studio/node_modules/@dolittle/vanir-backend/graphql/GraphQLSchema.ts:39:34)",
            "    at IOCContainer.getInstance (/Users/gh/code/dolittle/Studio/node_modules/type-graphql/dist/utils/container.js:39:26)",
            "    at /Users/gh/code/dolittle/Studio/node_modules/type-graphql/dist/resolvers/create.js:15:51",
            "    at field.resolve (/Users/gh/code/dolittle/Studio/node_modules/apollo-server-core/src/utils/schemaInstrumentation.ts:103:18)",
            "    at resolveField (/Users/gh/code/dolittle/Studio/node_modules/graphql/execution/execute.js:466:18)",
            "    at executeFields (/Users/gh/code/dolittle/Studio/node_modules/graphql/execution/execute.js:294:18)"
          ]
        }
      }
    }
  ],
  "data": {
    "entities": null
  }
}```

It should have given me a clue about the tenant-id missing from the header.

Here is my query

```
query {
  entities {
    allEntities {
      _id
    }
  }
}
```

Of course, this work very well when tenant id is added to the header. I suggest you fix this in Vanir, and then updated the Vanir dependency in this project.

**Comment by einari** [comment]
Sorry, my bad - I gave the wrong URL. It's in plural, so it should be `http://localhost:9000/_/businessmomentsadapters/graphql`

File: Source/BusinessMomentsAdapters/Backend/domain/miners/Miner.ts [hunk 3]
```diff
@@ -0,0 +1,18 @@
...
+
+import { AggregateRoot, aggregateRoot } from '@dolittle/sdk.aggregates';
+import { Guid } from '@dolittle/rudiments';
+import { EntityAssociatedWithMiner, ImperativeEmbedDefined } from '../../events/miners';
+
+@aggregateRoot('f4d99676-cd37-4e7e-8a03-95cd29b5927d')
+export class Miner extends AggregateRoot {
+
+    associateEntity(entityId: Guid) {
+        this.apply(new EntityAssociatedWithMiner(entityId.toString()));
```
goeran, line 12: Should we allow duplicate associations?

goeran (changes requested):
Work delivered with great speed. I have some questions and observations. Please answer before merge in.

Source/BusinessMomentsAdapters/Backend/domain/adapters/Adapter.ts
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) Dolittle. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+import { AggregateRoot, aggregateRoot, on } from '@dolittle/sdk.aggregates';
```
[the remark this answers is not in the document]
einari: Good point. I guess I'm so used to this, that it didn't even ring a bell :) 
In JS/TS, the naming convention that has become idiomatic is that things that can be instantiated, such as a class are with PascalCasing, while functions and things that can't be instantiated should be lowerCamelCase. Found a good blogpost on it: https://www.robinwieruch.de/javascript-naming-conventions

Casing matters and they are two different things, one being the decorator and the other a base class.

We have this for the C# SDK as well - and in a way worse in the sense that the name is exactly the same without case differences for the AggregateRoot attribute and AggregateRoot base class.
https://dolittle.io/docs/tutorials/aggregates/

Source/BusinessMomentsAdapters/Backend/domain/adapters/for_Adapter/when_adding_connector/and_there_is_an_existing_with_same_name.ts
```diff
@@ -0,0 +1,28 @@
...
+    const adapter = new Adapter(EventSourceId.from('46e44f22-1e76-4110-8b1d-d088cd3a1481'));
+    adapter.onConnectorAddedToAdapter({
+        name,
+        connectorId: 'a60bd940-7a3f-417d-bf2d-8126513fe3fc',
+        connectorTypeId: ConnectorTypes.WebHook.toString()
+    });
+
+    let brokenRuleError: BrokenRuleError;
+
+    try {
```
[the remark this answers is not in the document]
einari: The variable is outside the beforeEach - but its value is set in the beforeEach. If we had the declaration inside the beforeEach, we wouldn't have access to it in the `it` statements

Source/BusinessMomentsAdapters/Backend/domain/connectors/AddMinerToConnector.ts
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Dolittle. All rights reserved.
```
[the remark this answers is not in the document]
einari: This is a legal question with two aspects:

1. Clear copyright owner
2. Reference to license - and for us "use at own risk"

There has been quite some debate around this on the interwebs and it didn't seem clear what was best. From this we made a decision to keep these to be 100% sure. The debate is around whether or not the context of the repository is sufficient or if every artefact in the repository must mention this and refer to the license file.

To make life easier, I have created snippets for all languages in VSCode that is called 'dolittleLicense'. Whenever I start a new file, I just start typing `dolittle` and it gives me the header.

ESlint is configured to catch these, and right now we have disabled the automatic insertion of the license header on save. The reason for disabling it was that on occassion, it went bananas and inserted 50 license headers on save :)

Inside the `.vscode` folder sits the `settings.json` file that holds this:

```json
    "editor.codeActionsOnSave": {
        "source.fixAll.eslint": false
    },
```

We could enable this again and just ignore the few times it goes bananas - you do see it when it happens, and its a mere Ctrl/Cmd+Z moment when it happens.

Source/BusinessMomentsAdapters/Backend/read/entities/EntityQueries.ts
einari: Good point. I've added an issue for this in Vanir: https://github.com/dolittle-entropy/vanir/issues/192

Source/BusinessMomentsAdapters/Backend/domain/adapters/Adapter.ts
```diff
@@ -0,0 +1,39 @@
...
+
+@aggregateRoot('08dfe483-0b6a-40e1-95b7-1b3c7ccaf187')
+export class Adapter extends AggregateRoot {
+    private _connectors: string[] = [];
+
+    constructor(eventSourceId: EventSourceId) {
+        super(eventSourceId);
+    }
+
+    addConnector(name: string, connectorId: Guid, connectorTypeId: Guid): void {
```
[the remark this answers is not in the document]
einari: Very good point :)    We should not - I'll fix this.

Source/BusinessMomentsAdapters/Backend/domain/miners/Miner.ts
einari: Good question. 

From a domain perspective, we also need to decide if associations can only happen once and never changed.
If we allow for changing the entity association, I see two approaches to this: 

1. Introduce a rule that will break if you associate with the same entity twice
2. Silently ignore if one associates with the same entity twice (not apply an event)

Source/BusinessMomentsAdapters/Backend/domain/connectors/AddMinerToConnector.ts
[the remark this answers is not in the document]
einari: I found some guidance from the .NET foundation that might help: https://forums.dotnetfoundation.org/t/file-headers-and-copyright-statements/1276

Source/BusinessMomentsAdapters/Backend/domain/adapters/Adapter.ts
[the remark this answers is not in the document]
goeran: @einari Thanks. I just wanted to mention it, but it's out of scope for this PR.

Source/BusinessMomentsAdapters/Backend/domain/adapters/for_Adapter/when_adding_connector/and_there_is_an_existing_with_same_name.ts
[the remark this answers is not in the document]
goeran: @einari I agree regarding scoping the variable outside the beforeEach.

However, I can't see there is a `beforeEach` being invoked in this test? The way I interpret this code (in my head) is that adaptor.onConnectorAddedToAdapter() method invocation will be called just once - when the describe block is being invoked. If there was multiple it-blocks in this file the initialisation would only be executed once - hence the temporal coupling. If variables was initialised in an actual beforeEach the preconditions would have been the same for all the it-blocks.

Source/BusinessMomentsAdapters/Backend/domain/connectors/AddMinerToConnector.ts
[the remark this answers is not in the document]
goeran: @einari Thanks for clearing this up. Valuable for me to know. I think we should continue as-is.

Source/BusinessMomentsAdapters/Backend/domain/miners/Miner.ts
goeran: Hmmm...

I think we should avoid producing duplicate events when associating entity with miner. I don't see any benefits of the event being duplicated. Do you? I'm just asking in case I'm ignoring some details.

I don't think it's a good solution to silently ignore duplicated associations. The rule will give a very clear hint to the developer that the entity has already been associated. This can help when debugging on the consumer side.

Even though we don't have a use case for it yet, I have a feeling that we'll need to be able to remove associations too.

I think we should go for option 1.

Source/BusinessMomentsAdapters/Backend/domain/miners/Miner.ts
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Dolittle. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+import { AggregateRoot, aggregateRoot } from '@dolittle/sdk.aggregates';
+import { Guid } from '@dolittle/rudiments';
+import { EntityAssociatedWithMiner, ImperativeEmbedDefined } from '../../events/miners';
+
+@aggregateRoot('f4d99676-cd37-4e7e-8a03-95cd29b5927d')
+export class Miner extends AggregateRoot {
```
goeran, line 9: I have a question about how we name aggregates.

I think Miner is a good name, but looking on it from the outside it isn't clear to me that this is an Aggregate. Type info about the base class gives me a hint, and when reading the implementation of the class it's super clear to me that this is an aggregate. However, I need to check type info or read the implementation to see that this is an aggregate.

I like to use fuzzy search to find files. If this was named MinerAggregate it would be easy for me to find the aggregates in our solution. For the command handlers for the Adapter we use this convention of naming - see `AdapterCommandHandlers.ts`. 

@einari

Source/BusinessMomentsAdapters/Backend/domain/entities/Entity.ts
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Dolittle. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+import { AggregateRoot, aggregateRoot } from '@dolittle/sdk.aggregates';
+import * as events from '../../events/entities';
+
+@aggregateRoot('4687e8dd-95fa-4d01-8ef1-5fa4644acb5c')
+export class Entity extends AggregateRoot {
+    defineFilter(code: string): void {
```
goeran, line 9: I believe we can have multiple entities? If that's the case we should probably link the filter, transform and projection to that id?

goeran: @einari 👆

Source/BusinessMomentsAdapters/Backend/public/index.html
```diff
@@ -0,0 +1,9 @@
+<html>
```
goeran, line 1: Is this file needed? @einari

Source/BusinessMomentsAdapters/Backend/read/entities/EntityQueries.ts
goeran: @einari  nice!

Source/BusinessMomentsAdapters/Backend/domain/adapters/for_Adapter/when_adding_connector/and_there_is_an_existing_with_same_name.ts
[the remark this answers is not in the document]
einari: :) Sorry, I was completely out of context - and assumed there was a beforeEach..  

Good point about the temporal coupling.
I don't necessarily put in beforeEach for all specs. I don't think for this case there would be any benefit in having a beforeEach now. Maybe if I were adding another it - but that would be depend on what I was doing. 

One could argue we should always do it, to avoid potentially having problems. Which would be fine by me.

Source/BusinessMomentsAdapters/Backend/domain/entities/Entity.ts
einari: Could you elaborate a bit on this? There can be multiple of the type `Entity`. What I see however is that it lacks a constructor for taking the EventSourceId and pass it along to the base class - which is missing. A bit of a pity that TypeScript doesn't warn about this on compile step. The base class of `AggregateRoot` has a constructor that requires this - and it would be extremely useful if we got a compiler error for that.

I'll put in the constructor.

Source/BusinessMomentsAdapters/Backend/domain/miners/Miner.ts
einari: I realize I'm very inconsistent here.

The thinking is that in the context of working on the domain, you wouldn't by accident use a read model that would happen to have the same name. The thinking has been that structure and context should be enough to know what you're working on. We have principles on naming described here: https://dolittle.io/docs/contributing/guidelines/naming/#prefix--postfix

When searching for instance with ctrl/cmd+P in VSCode you'd see the context and structure it is part of:
image

But I guess if you're searching for which aggregate roots you have available, that won't help you :) 

Suggest we have a chat about this - the codebase is extremely inconsistent on it - and in fact, most of the constructs postfixes and completely violates the principle :)
Question then would be; would it make sense to strip the postfixes and have clearer names or go the other way and be consistent on postfixes. 

When we put the principle in place, one of the thinkings where that with a pre/post fix you create a conceptual coupling and it is a bit of a "cop out" for naming things properly. But there are nuances to this.

In my head, I don't think about the technical artifact, but what it is doing rather. For instance commands & events, a command through its name represents its intent clearly, not that it is a command - and same for an event, it represents what has happened - not that it is a technical thing of "Event".

Source/BusinessMomentsAdapters/Backend/public/index.html
einari: To be honest, I can't remember why its there and if it is needed :) 
When we containerize it - the backend serves the static content through an Express middleware and it is set to look for things in the public folder. 

This could be the result of an overprotective strategy of letting the dev know if they point their browser to the backend during development that they're looking in the wrong place instead of getting a 404 and spending time trying to figure out what is giving this.

Source/BusinessMomentsAdapters/Backend/domain/miners/Miner.ts
einari: There could be one benefit to using postfixes; cohesion. 

Right now, we have a structure of `./domain` `./read`in the root of the project and then feature folders within these.
If we postfix `Miner` with `MinerAggregate` or prefix it with `AggregateOfMiner` and then the `MinerReadModel`, we could keep all artifacts related to the same feature for the tier (backend) in the same feature folder and not split out on concern like now. #FoodForThought

PR #70, opened 2021-04-22
Business moments adaptor UI

goeran:
# Todo
- [ ] How do we get the domain prefilled into the webhook configuartion screen?

# How to run
## Start mongo
```
cd .dolittle
docker-compose up -d studio-mongo
```

### Make sure yarn is happy
```
yarn
```

###  Start frontend
- currently doesnt need a backend
```
cd Source/BusinessMomentsAdapters/Web
yarn start:dev
```

### Open Browser
Goto http://localhost:9006/


freshteapot:
I have some ugly UI's that is helping learn FluentUI and has monaco editor thrown in.
The breadcrumb bar is only there to make it easier to jump around

einari:
Looking great. 

Input on composition, routing and navigation. 
The Studio composition is based on an iframe model, described here:
https://github.com/dolittle-entropy/vanir/blob/main/Documentation/frontend/composition.md

The goal is to have the commonalities in the Portal, such as: 
- Navigation structure (Top level menu - right side menu)
- Breadcrumbs
- Toolbars
- Notifications

I noticed you have put in Breadcrumbs for instance. Which is something we're missing in the Portal right now, but would be great to move into there to get that consistently for all microservices.

With the model we've been aiming for also having deeplinking from the top Window based on the conventions of mapping the <microservicename> configured individually in each of the microservices `vanir.json` files - reflected in the local environment setup of the Nginx ingress. This would give us `https://dolittle.studio/businessmomentsadapters` at the top level, which for the `iframe` inside the composition would become `https://dolittle.studio/_/businessmomentsadapters`. Anything after that in the URL would then be appended to the `iframe` URL. 

In order for this to work, there are some constructs in Vanir that helps with this. I see you've gone ahead and used `<BrowserRouter/>` for instance from the React Router. There are wrappers for this that will adhere to the configuration and conventions put in place:
https://github.com/dolittle-entropy/vanir/blob/main/Documentation/frontend/react/routing.md

Finally, you can then also do navigation from a View Model, if needed - leveraging the navigator.
https://github.com/dolittle-entropy/vanir/blob/main/Documentation/frontend/navigator.md


While working with local development, I recommend running with the portal and through the portal. Then you get the correct look and feel and would see it all fitting together.

Inside the Shared folder you'll see `Portal` and `Components`. Portal contains APIs for working with some of the things found at the Portal level, such as the Toolbar and Navigation. This probably needs some documentation in the repo.

As an example of how this all fits together, you can either go to the previous implementation of the `Applications` microservice found in the Development branch - or look at the new implementation, now called `Deployables` in the draft PR #68

Source/BusinessMomentsAdapters/Web/configuration/BasicAuthComponent.tsx
```diff
@@ -0,0 +1,25 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+import React from 'react';
+import { TextField, ITextFieldStyles } from '@fluentui/react/lib/TextField';
+import { Label } from '@fluentui/react/lib/Label';
+import { Stack } from '@fluentui/react/lib/Stack';
+
+const textFieldStyles: Partial<ITextFieldStyles> = { fieldGroup: { width: 300 } };
+const stackTokens = { childrenGap: 15 };
+
+export const BasicAuthComponent: React.FunctionComponent = () => {
```
einari, line 11: You don't need to include the type `React.FunctionComponent`.

```suggestion
export const BasicAuthComponent = () => {
```

Source/BusinessMomentsAdapters/Web/configuration/WebhooksConfiguration.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { IDropdownOption } from '@fluentui/react';
+import { PrimaryButton } from '@fluentui/react/lib/Button';
+
+import { BasicAuthComponent } from './BasicAuthComponent';
+import { BearerAuthComponent } from './BearerAuthComponent';
+
+const textFieldStyles: Partial<ITextFieldStyles> = { fieldGroup: { width: 300 } };
+const stackTokens = { childrenGap: 15 };
+
+export const WebhooksConfig: React.FunctionComponent = () => {
```
einari, line 17: Can be simplified:
```suggestion
export const WebhooksConfig = () => {
```

einari (commented):
We should probably consolidate on structure - to keep with the structure and cohesion principles: https://dolittle.io/docs/contributing/guidelines/development_principles/#high-cohesion

Comment to the `Store` - related to the cohesion comment above.
The ViewModel (https://github.com/dolittle-entropy/vanir/blob/main/Documentation/frontend/react/view-models.md) gives a way to keep it separated from the View - and maintain cohesion. 

For those scenarios when a ViewModel is not needed and you'd rather just have it in the view, the `useQuery()` and `useMutation()` hooks from Apollo would be nice. I've added an issue on Vanir to make sure we can do that and it leveraging the same configuration (https://github.com/dolittle-entropy/vanir/issues/194). For more details on Apollo: https://www.apollographql.com/docs/react/data/queries/

Something that would be nice then is to have the query in a separate file:

`allConnectorsQuery.ts`:
```typescript
import gql from 'graphql-tag';

export const allConnectorsQuery = gql`
    query {
        connectors {
             allConnectors {
                  id,
                  name
             }
        }
    }
`;
```

PR #71, opened 2021-04-30
The Business Moments Adaptor Docker Image

goeran:
The actual image that will be used to create Business Moments Adaptor microservices (pods with containers).

# How to build docker image
```
# be in the root folder
docker build -t businessmomentsadaptor:latest Source/BusinessMomentsAdaptorImage
```


Source/BusinessMomentsAdaptorImage/server.js
```diff
@@ -0,0 +1,28 @@
+// Require express and body-parser
```
goeran, line 1: @parhambarazesh It seems to me that this js file can be delete now as the typescript file is now working?

Source/BusinessMomentsAdaptorImage/index.ts
```diff
@@ -0,0 +1,29 @@
...
+const app = express();
+const PORT = 3008;
+
+// Tell express to use body-parser's JSON parsing
+
+app.use(bodyParser.json());
+
+app.post('/api/webhooks-ingestor', (req: any, res: any) => {
+    console.log(req.headers.authorization);
+    if (req.headers.authorization.split(' ')[1] === process.env.WH_TOKEN) {
```
goeran, line 20: If no auth header is provided in the request it will cause an exception, because the split method (req.headers.authorization.split) will return an empty array.

goeran: I suggest you drop the split and expect the type of auth header to be provided as environment variable `Basic bTM6am9obmNhcm1hY2s=`. This will also make the code cleaner.

Source/BusinessMomentsAdaptorImage/index.ts
```diff
@@ -0,0 +1,29 @@
...
+// Initialize express and define a port
+const app = express();
+const PORT = 3008;
+
+// Tell express to use body-parser's JSON parsing
+
+app.use(bodyParser.json());
+
+app.post('/api/webhooks-ingestor', (req: any, res: any) => {
+    console.log(req.headers.authorization);
```
goeran, line 19: Remove the console log.

package.json
```diff
@@ -3,7 +3,9 @@
     "workspaces": [
         "Source/**/*"
     ],
-    "dependencies": {},
+    "dependencies": {
+        "atob": "^2.1.2"
```
goeran, line 7: It seems to me that this is no longer in use, and can be removed?

goeran (commented):
I added some comments. Nice first stab at this, and this is soon in a state that we can use it for the demo on Friday. Good work 👏

Source/BusinessMomentsAdaptorImage/server.js
parhambarazesh: Not sure yet. If we remove server.js, then should I include the "tsc server.js" command in the docker file? or creating the js file is handled somehow in the pipeline?

Source/BusinessMomentsAdaptorImage/index.ts
parhambarazesh: I have changed the env variable to "Basic bTM6am9obmNhcm1hY2s=". but the type in ts code is a string. Do we have a type representing the env variable?

Source/BusinessMomentsAdaptorImage/index.ts
parhambarazesh: Done!

Source/BusinessMomentsAdaptorImage/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "businessmomentsadaptorimage",
+  "version": "1.0.0",
+  "description": "",
+  "main": "server.js",
+  "scripts": {
+    "test": "NODE_ENV=test mocha --check-leaks -r tsconfig-paths/register -r ts-node/register \"**/for_*/**/*.ts\"",
```
parhambarazesh, line 7: For an unknown reason to me, I need to pass the NODE_ENV variable and check if the code is in the test scenario or not. It seems during the test, the code is trying to listen to the same port twice and I will get a "port already in use" error. While I expect it to run only once and the error has no sense to me. I have solved it in this way, by making the server.ts not listen to the port while testing. Think there should be better ways to solve the issue.
Any comments?

PR #73, opened 2021-05-18
Store raw data for incoming webhooks

parhambarazesh:
https://app.asana.com/0/1200313386408105/1200309624168622

# How to test?

To run the code cd to Source/BusinessMomentsAdaptorImage and run:
yarn ts-node index.ts
Make sure you have installed the packages; like mongoose.


Source/BusinessMomentsAdaptorImage/dataController.ts
```diff
@@ -0,0 +1,33 @@
+import { Request, Response } from 'express';
```
goeran, line 1: Add copyright

```
// Copyright (c) Dolittle. All rights reserved.
// Licensed under the MIT license. See LICENSE file in the project root for full license information.

```

Source/BusinessMomentsAdaptorImage/dataController.ts
```diff
@@ -0,0 +1,33 @@
...
+import rawData from './dataschema';
+
+export let allData = (req: any, res: any) => {
```
goeran, line 4: allData is never reassigned. Use `const` instead.

You can run the following command to get the linting feedback.

```bash
yarn ci
```

Source/BusinessMomentsAdaptorImage/dataController.ts
```diff
@@ -0,0 +1,33 @@
...
+    let data = rawData.find((err: any, data: any) => {
```
goeran, line 5: Use `const` instead of `let`. The variable named `data` is never reassigned.

Source/BusinessMomentsAdaptorImage/dataController.ts
```diff
@@ -0,0 +1,33 @@
...
+        if (err) {
+            res.send('Error!');
+        } else {
+            res.send(data);
+        }
+    });
+};
+
+export let getData = (req: Request, res: Response) => {
```
goeran, line 14: Use const.

Source/BusinessMomentsAdaptorImage/dataschema.ts
```diff
@@ -0,0 +1,25 @@
+import mongoose from 'mongoose';
+
+const uri: string = 'mongodb://127.0.0.1:27017/local';
```
goeran, line 3: This URI has to be configurable - use ENV var to configure this.

Source/BusinessMomentsAdaptorImage/dataController.ts
```diff
@@ -0,0 +1,33 @@
...
+    rawData.findById(req.params.id, (err: any, data: any) => {
+        if (err) {
+            res.send(err);
+        } else {
+            res.send(data);
+        }
+    });
+};
+
+export let addData = (req: Request, res: Response) => {
```
goeran, line 24: use const.

Source/BusinessMomentsAdaptorImage/dataController.ts
```diff
@@ -0,0 +1,33 @@
...
+    let data = new rawData(req.body);
```
goeran, line 25: use const.

Source/BusinessMomentsAdaptorImage/dataschema.ts
```diff
@@ -0,0 +1,25 @@
...
+
+mongoose.connect(uri, (err: any) => {
+    if (err) {
+        console.log(err.message);
+    } else {
+        console.log('Successfully Connected to MongoDB!');
+    }
+});
+
+export const DataSchema = new mongoose.Schema({
```
goeran, line 13: I don't think we can use schema for the payload. If the payload changes we'll break the schema. Try to rename some of the properties - like publisher. What happens? My bet is that the publisher field will not be stored - it will be ignored.

Source/BusinessMomentsAdaptorImage/dataschema.ts
```diff
@@ -0,0 +1,25 @@
...
+    publisher: { type: String, required: false },
+    operation: { type: String, required: false },
+    document: { type: String, required: false },
+    timestamp: { type: String, required: false },
+    id: { type: String, required: false },
+    elements: { type: Array, required: false },
+    message: { type: String, required: false },
+});
+
+const rawData = mongoose.model('rawData', DataSchema);
```
goeran, line 24: Here is how you can fix the schema issue,

```
const rawData = mongoose.model('rawData', new mongoose.Schema({ any: {} }, { strict: false }));

```

Source/BusinessMomentsAdaptorImage/server.ts
```diff
@@ -1,36 +1,45 @@
...
 
     app.post('/api/webhooks-ingestor', (req: Request, res: Response) => {
         if (req.headers.authorization === process.env.WH_AUTHORIZATION) {
             console.log(req.body);
+
+            let data = new rawData(req.body);
+            try {
+                data.save();
+            } catch (error) {
+                res.send(error);
```
goeran, line 23: Let's not return the error in the response. If we get an exception we can log it and return the HTTP status code for the internal server error.

We don't know what the error msg might unveil.

goeran (changes requested):
Hi @parhambarazesh. I've made some comments.

goeran:
@parhambarazesh I think you can define TTL on the schema level for mongoose: https://mongoosejs.com/docs/api.html#schematype_SchemaType-index

Source/BusinessMomentsAdaptorImage/dataController.ts
parhambarazesh: Fixed

Source/BusinessMomentsAdaptorImage/dataController.ts
parhambarazesh: Fixed

Source/BusinessMomentsAdaptorImage/dataController.ts
parhambarazesh: Fixed

Source/BusinessMomentsAdaptorImage/dataschema.ts
parhambarazesh: I agree and I was thinking of that before. But I was not sure how to make a generic schema to handle any type of payload. And yes, in case you have a different payload (the keys), the key will not be stored. Needs to be fixed

Source/BusinessMomentsAdaptorImage/dataschema.ts
parhambarazesh: Seems this is not working. I will have more dig into it to make a generic schema.

parhambarazesh: My mistake. It is working!

Source/BusinessMomentsAdaptorImage/dataController.ts
parhambarazesh: Fixed

Source/BusinessMomentsAdaptorImage/dataController.ts
parhambarazesh: FIxed

PR #74, opened 2021-05-19
Transpile typescript to js

goeran:
With this PR we can transpile typescript to js and use node and js inside the container to run our code. We don't need to install typescript in the container.

### How to build

```
# be in root folder
docker build -t businessmomentsadaptor:latest . -f Source/BusinessMomentsAdaptorImage/Dockerfile
```


Source/BusinessMomentsAdaptorImage/Dockerfile
```diff
@@ -1,24 +1,44 @@
...
+# Resulting Image
+FROM node:14.2.0-slim as base
+ARG CONFIGURATION=Release
+
+RUN echo Configuration = $CONFIGURATION
+
+WORKDIR /app
+COPY --from=backend-build /app/Source/BusinessMomentsAdaptorImage/dist/index.js ./
+
+ENTRYPOINT ["node", "index.js"]
```
parhambarazesh, line 44: I checked this PR. It seems good. but can you explain how we can benefit from having js in the container? I supposed we would like to have TS as the base language in codes/containers. Maybe we can remove typescript installation from the dockerfile, with ts-node we still can be able to convert the TS into JS inside the container. Have you changed the dockerfile to use JS as the base language, in order to make the containers smaller/less complicated?

Source/BusinessMomentsAdaptorImage/Dockerfile
```diff
@@ -1,24 +1,44 @@
...
+COPY Source/Shared/Web/package.json ./Source/Shared/Web/package.json
 RUN yarn
 
-# If you are building your code for production
-# RUN npm ci --only=production
+#COPY Source/BusinessMomentsAdaptorImage/Backend/microservice.json /app/Source/BusinessMomentsAdaptorImage/microservice.json
 
-# Bundle app source
-COPY . .
+# Backend Build
```
parhambarazesh, line 19: Do we need to set node base image twice?

Source/BusinessMomentsAdaptorImage/Dockerfile
```diff
@@ -1,24 +1,44 @@
...
+COPY Source/Shared/Backend/package.json ./Source/Shared/Backend/package.json
+COPY Source/Shared/Common/package.json ./Source/Shared/Common/package.json
+COPY Source/Shared/Specs/package.json ./Source/Shared/Specs/package.json
+COPY Source/Shared/Styles/package.json ./Source/Shared/Styles/package.json
+COPY Source/Shared/Web/package.json ./Source/Shared/Web/package.json
 RUN yarn
 
-# If you are building your code for production
-# RUN npm ci --only=production
+#COPY Source/BusinessMomentsAdaptorImage/Backend/microservice.json /app/Source/BusinessMomentsAdaptorImage/microservice.json
```
parhambarazesh, line 17: Some unused commented lines. Good to remove them.

Source/BusinessMomentsAdaptorImage/Dockerfile
```diff
@@ -1,24 +1,44 @@
...
-# where available (npm@5+)
 COPY package.json ./
+COPY .eslintrc.js ./
+COPY tsconfig.json ./
+#COPY application.json ./
 
-RUN npm install -g typescript ts-node
-#RUN yarn add typescript
-#RUN yarn add ts-node
+COPY Source/Shared/Backend/package.json ./Source/Shared/Backend/package.json
```
parhambarazesh, line 10: What is the purpose of copying these config files?

goeran: @parhambarazesh to get some of the standard packages we use.

Source/BusinessMomentsAdaptorImage/Dockerfile
goeran: @parhambarazesh You are right. I should remove them.

Source/BusinessMomentsAdaptorImage/Dockerfile
goeran: Typescript is transpiled to JavaScript and then executed by nodejs. I'm not sure what you mean by suggesting to remove typescript install from the dockerfile. Can you elaborate? 

Yes, I have changed to use js files to make the container smaller.

parhambarazesh: It's a good point to have smaller containers.
I supposed you were pointing to installing typescript inside the old dockerfile, when you said we don't need to install it. I meant we could remove it and leave the installation of ts-node as it works.
But in general, that makes sense to have the containers based on JS.

Source/BusinessMomentsAdaptorImage/Dockerfile
parhambarazesh: I have seen a similar structure in the other dockerfiles. That looks meaningful to me. Though I'm not sure how we are going to use these packages. As far as I know, they are not used within the same microservice. Maybe we need them elsewhere?

PR #75, opened 2021-05-19
More tests for BusinessMomentsAdaptors webhooks

goeran:
What's done?

- added more tests
- refactored the tests to make them more readable
- fixed linting errors

NB: We should merge #74 before this and then change base for this PR.


Source/BusinessMomentsAdaptorImage/for_server/when_post_webhook_and_not_authorized.ts
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) Dolittle. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+import supertest from 'supertest';
```
parhambarazesh, line 4: In the previous file, You have imported "request" from supertest package. This time you import supertest. But it seems they have a similar structure. Any reason to use different methods from supertest package?

PR #77, opened 2021-05-25
Post buisness moments

freshteapot:
# What
- Fix creating business moments via ui
- Make delete work


freshteapot (commented):
Good enough

PR #80, opened 2021-06-03
Adding SDK and runtime config to Business Moments Adaptor

parhambarazesh:
PR for adding SDK and runtime config to Business Moments Adaptor.

Solvs https://app.asana.com/0/1200403658239260/1200358538105062/f

### DIFF
--- a/Source/BusinessMomentsAdaptorImage/server.ts
+++ b/Source/BusinessMomentsAdaptorImage/server.ts
@@ -6,6 +6,11 @@ import bodyParser from 'body-parser';
 import process from 'process';
 import { IRawDataStorage } from './RawDataStorage';
 
+import { Client } from '@dolittle/sdk';
+import { TenantId } from '@dolittle/sdk.execution';
+import { WebhookReceived } from './WebhookReceived';
+import { WebhookReceivedHandler } from './wehbookReceivedHandler';
+
 export function createServer(rawDataStorage: IRawDataStorage) {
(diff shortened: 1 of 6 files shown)


Source/BusinessMomentsAdaptorImage/server.ts
```diff
@@ -23,6 +27,14 @@ export function createServer(rawDataStorage: IRawDataStorage) {
         try {
             await repo.Append(req.body);
             console.log(req.body);
+            const client = Client.forMicroservice('a87b3411-7f23-4ea3-9b74-f6fc7f7c5481')
```
goeran, line 30: The client object will now be built every time the webhook request is called. I suggest you make this in index.ts and send it in as an parameter in the createServer function

Source/BusinessMomentsAdaptorImage/webhookReceived.ts
```diff
@@ -0,0 +1,6 @@
+import { eventType } from '@dolittle/sdk.events';
+
+@eventType('0cb79cbf-bbe7-445c-99fa-0db6f7f24d4f')
```
goeran, line 3: Thanks for providing this example event.

goeran (changes requested):
I had one suggestion for change. You can make that change and then merge in this PR.

goeran (commented):
@parhambarazesh This now looks good!

PR #86, opened 2021-06-09
Raw Data Log Webhook UI Part 1

freshteapot:
# Todo
- [x] Add Create microservice for raw data log
- [x] Post to platform-api
- [x] Make platform-api work with new microservice
- [x] Add Config screen for raw data log
- [x] Add "Add new Webhook" to the Webhooks screen in raw data log

# Reference
- https://app.asana.com/0/1200403658239260/1200420198454048/f


freshteapot (commented):
Good enough, leaving the todos as is, to keep making small changes.

PR #89, opened 2021-08-24
Renovation

woksin:
# What
- Delete unused code

Resolves https://app.asana.com/0/0/1200834923578009/f


freshteapot (approved):
Great work @woksin.

I focused on the "additions" or "changes", and less on the red / deletion code.

I confirmed it was happy running locally from an empty node_modules.
It is also running in dev.dolittle.studio.

PR #91, opened 2021-08-25
Fix SelfService Web Dockerfile

woksin:
## Summary

Fix the Dockerfile for SelfService Web

### Fixed

- The Dockerfile for SelfService Web

### DIFF
--- a/Source/SelfService/Web/Dockerfile
+++ b/Source/SelfService/Web/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:14.17.5 AS shared-build
+FROM node:16.7.0 AS shared-build
 
 # Install dependencies
 WORKDIR /build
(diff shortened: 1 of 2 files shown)


freshteapot:
@woksin I added some docs about manually running it via docker (I also wanted to test the dockerfile).

The action is not happy (perhaps you know this)
https://github.com/dolittle/Studio/actions/runs/1166038495

Source/SelfService/Web/Dockerfile
```diff
@@ -1,45 +1,26 @@
-FROM node:14.2.0 AS shared-build
+FROM node:14.17.5 AS shared-build
```
freshteapot, line 1: What if we go crazy and use a more modren node :)

```suggestion
FROM node:16.7.0 AS shared-build
```

> node --version
v16.7.0

woksin:
I know, that's because there are no actions here. The other PR i have will fix this

freshteapot (approved):
Good stuff, nice work.

PR #105, opened 2021-09-10
Add mongo to the local dev setup

joelhoisko:
## Summary
Adds a hardcoded mongo deployment & service to the local dev setup. As Studio doesn't yet dynamically create the required mongodb this is required to get the local dev setup bootstrapped.

Remember to remove sections that you don't need or use.

### Added

- New `mongo.yml` to be applied during the bootstrapping phase
- Docs on how to apply it & why we do bootstrapping

### DIFF
--- a/Documentation/k3d.md
+++ b/Documentation/k3d.md
@@ -44,13 +44,12 @@ Do the following command and verify that current context is `k3d-dolittle-env`
 kubectl config current-context
 ```
 
-Execute the following commands in the `k3d-dolittle-dev` context:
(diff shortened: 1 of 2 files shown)


joelhoisko:
I tried it locally, and the mongo is all booted up but as the "Base (default)" microservice from Studio doesn't do anything with events I'm not 100% that it works. I checked the `resources.json` for the "Base (default)" microservice and it had the correct mongo connection strings.

One thing that I didn't add was backups, as I wasn't on how we wanna deal with them. Wouldn't be nice to spam the users local computer with a backup every hour etc.

jakhog (approved):
LGTM. Agreed with holding off with backups (and persisting state for Mongo) - we don't really need it now.

PR #108, opened 2021-09-20
Fix the COPY TO CLIPBOARD button to actually work in POAPI creation view

joelhoisko:
## Summary

Fixes the "COPY TO CLIPBOARD" button in the PurchaseOrderAPI microservice creation view to actually work.

### Fixed

- The "COPY TO CLIPBOARD" button actually works now


joelhoisko:
One thing that I didn't deal with was that the `navigator.clipboard.writeText()` actually [returns a Promise](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText), but we don't do anything with the result right now so I just left it like that. In the future we could have a "URL copied" text popup or something if we want.

jakhog (approved):

jakhog:
I might have a look at the notification (need to look at the preferred Material Design way to do it) - but as you said, for a later PR :)

PR #110, opened 2021-09-20
Bug fix white name++

freshteapot:
## Summary
- Fixing the white text for name
- Fixing the white text for username and password
- Removing the warning messages in the console
- Making the file names make a little more sense
- Delete icon and text is white
- Validation on username and password
- Validation on name of microservice

# Reference
- https://app.asana.com/0/1200899594200551/1200948921136344/f


joelhoisko (changes requested):
Nice stuff, just one point about validation. I also saw that in the creation view the inactive steps text is super dark and hard to see:

image

Looking at the Firma design it should be more greyish and easily visible:

image


I'm not sure if this is the PR that also touches upon that text color or is it hidden in some other ticket, this one just sounded the closest one to fixing this thing too.

freshteapot:
@joelhoisko the "tick" is wrong on completed (i can live with that), but the rest is closer.

joelhoisko (approved):

PR #114, opened 2021-09-22
Remove the views that allow creation of a RawDataLog

jakhog:
## Summary

Removes the views that allow creation of a Raw Data Log Ingestor manually. It is created automatically when the user creates a PurchaseOrderAPI instead.

### Removed

- The ability to manually create a Raw Data Log Ingestor


freshteapot (approved):
Approved.

I agree with your approach, I was going to comment it out :P. But dead code is dead code.

I do hope to come to a day where a platform engineer has a create view where we can make more building blocks than the ones the end user can see.

I will make a little note of this PR.

This PR, now makes me wonder how we want to handle the delete button on the "viewCard" for the raw-data-log. (I will make a ticket)

jakhog:
@freshteapot: I might be more liberal on the delete button than others, so I'm happy to change there. I always keep thinking that we can just restore it from the history later, but then again - you might forget it's even there...

PR #116, opened 2021-09-24
Hide rawdatalog delete

woksin:
## Summary

Hides the delete button for the raw data log view card

#### How to test
- Start with a clean cluster (or without raw-data-log and purchase-order)
- Click Microservices
- Click Create new microservice
- Click Create within the Purchase order api card
- Select Infor graphic
- Enter name
- Click next
- Enter username
- Enter password
- Click Next
- Click Microservices
- See that there is no delete button on the raw-data-log-v1 webhook microservice 

### Added

- canDelete property to microservice view card

### Changed

- Uses HTML5 hidden to hide delete button from raw data log microservice view card

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,3 +12,11 @@ develop-backend:
 	DEVELOPMENT_TENANT_ID="453e04a7-4f9d-42f2-b36c-d51fa2c83fa3" \
 	DEVELOPMENT_USER_ID="local-dev" \
 	go run main.go
+
+clean-local-dev:
+	kubectl delete all --all -n application-11b6cf47-5d9f-438f-8116-0d9828654657
+	rm -r /tmp/dolittle-local-dev || true
+	mkdir -p /tmp/dolittle-local-dev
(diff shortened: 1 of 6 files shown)


joelhoisko (changes requested):
Found a bug:

I followed your steps, and after creating the POAPI I get greeted by the usual error screen. Then I click'd on "MICROSERVICES" from the left nav to get to the overview page and raw-data-log-v1 still had a Delete button:
image

If I refresh the page the button dissappears:
image

So somehow the `kind` is not set properly for the rawdatalog ms upon going to the overview view on the first go. Or something like that.

woksin:
Ohhh nice catch 😁

woksin:
The way I solved this led me into a bug in platform-api that I fixed here https://github.com/dolittle/platform-api/pull/35

joelhoisko (approved):
Tried with the bugfix from https://github.com/dolittle/platform-api/pull/35 and works fine 👍

PR #117, opened 2021-09-24
Add error handling to POAPI creation and display error with snackbar

joelhoisko:
## Summary

Handle PurchaseOrderAPI creation errors by displaying an error snackbar with the message from the HTTP response.

Preview (some of the HTTP responses are fixed in https://github.com/dolittle/platform-api/pull/34 as shown here):
image

### Added

- Error handling to POAPI creation view


Source/SelfService/Web/api/api.ts
```diff
@@ -156,9 +156,9 @@ export async function deleteMicroservice(applicationId: string, environment: str
     return result.status === 200;
 }
 
-export async function saveMicroservice(input: any): Promise<boolean> {
+export async function saveMicroservice(input: any): Promise<Response> {
```
jakhog, line 159: I think it would make a little sense do to more of the work here in checking that everything went okay.

What you could do is `await` the fetch call to get a response, then check that response if the `status code == 200`, and if not throw an error (could try to parse the body as json or just take the body text as the error).

If it was 200, then return the result after calling `.json()` on it, that would make this method return a nice result (seems like it is som actual good content there) or throw an error. What do you think?

Source/SelfService/Web/stores/microservice.ts
```diff
@@ -153,9 +153,9 @@ const saveMicroservice = async (kind: string, input: any): Promise<boolean> => {
             return Promise.resolve(false as boolean);
     }
 
-    if (!response) {
-        alert('Sad times');
-        return Promise.resolve(false);
+    const jsonResponse = await response.json();
```
jakhog, line 156: See my comment about doing this in the other function.

Source/SelfService/Web/stores/microservice.ts
```diff
@@ -153,9 +153,9 @@ const saveMicroservice = async (kind: string, input: any): Promise<boolean> => {
...
+    if (!response.ok) {
+        return Promise.reject(jsonResponse.message);
```
jakhog, line 158: I see that this was already done before you some places in this function, but since this function is `async` there is no need to wrap the return values in Promises, the compiler does that for you. For this code, the return type is actually a `Promise<Promise<any>>`.

jakhog (changes requested):
I have made some suggestions in comments.

Source/SelfService/Web/api/api.ts
joelhoisko: Changed it to throw an exception with the response message as the exception message

Source/SelfService/Web/stores/microservice.ts
joelhoisko: Removed the extra `Promise` stuff

Source/SelfService/Web/api/api.ts
```diff
@@ -168,8 +169,11 @@ export async function saveMicroservice(input: any): Promise<boolean> {
                 'content-type': 'application/json'
             }
         });
-    const jsonResult = await result.json();
-    return true;
+    const jsonResponse = await response.json();
```
freshteapot, line 172: It might be better to do this after the response.ok, as there is a chance there is no "json()".

Source/SelfService/Web/api/api.ts
```diff
@@ -156,9 +157,9 @@ export async function deleteMicroservice(applicationId: string, environment: str
     return result.status === 200;
 }
 
-export async function saveMicroservice(input: any): Promise<boolean> {
+export async function saveMicroservice(input: any): Promise<any> {
```
jakhog, line 160: Don't we have a type for the response in this case?

joelhoisko: There's no generic represenation of the returned data from platform-api. `MicroserviceSimple` with the `MicroserviceSimpleExtra` types are mostly there, but they don't have all of the properties that we get back for a POAPI creation, like `webhooks` or `writeTo`.

joelhoisko: I could make it into an `object` though

Source/SelfService/Web/api/api.ts
joelhoisko: I'll add a case for that within the `if (!response.ok)`. This case could happen if there is no body to parse into JSON so as to get the message from.

joelhoisko: Changed it now @freshteapot

Source/SelfService/Web/api/api.ts
joelhoisko: Also just as an observation, right now the response isn't being used for anything except showing the potential error message. It's just being passed through a few different functions until it lands in the original caller `_onSave` in `container.tsx` and it's just discarded.

jakhog: Ok, an object doesn't really help. I thought we had a matching TS declaration of the type - but then we can leave it for when we need the output later.

Source/SelfService/Web/stores/microservice.ts
```diff
@@ -132,7 +132,7 @@ export async function deleteMicroservice(
     return response;
 }
 
-const saveMicroservice = async (kind: string, input: any): Promise<boolean> => {
+const saveMicroservice = async (kind: string, input: any): Promise<any> => {
```
jakhog, line 135: If we don't use the actual data for anything, this should still be a boolean.

It feels a bit confusing if it returns either the data for a microservice, or false.

Source/SelfService/Web/stores/microservice.ts
```diff
@@ -153,11 +153,6 @@ const saveMicroservice = async (kind: string, input: any): Promise<boolean> => {
             return Promise.resolve(false as boolean);
```
jakhog, line 153:
```suggestion
            return false;
```

jakhog (changes requested):
Some responses in the comments, otherwise I think it would be easier to return true/false in the final evaluation if the saving worked (since we don't use the return data for anything). Let me know if that is confusing - I see how it could be based on my previous change request.

Source/SelfService/Web/stores/microservice.ts
joelhoisko: Changed it back to boolean, any potential errors are still handled through exceptions so that it bubbles up all the way to the frontend and snackbar. In the case of it returning `false` I made a basic snackbar error to appear.

jakhog:
@joelhoisko when testing this out locally I'm currently getting something that looks like this:
image

Is there something not merged into platform-api that causes it, or could it be a bug here?

jakhog:
Seems like the platform-api is returning a bit too much stuff:
image

jakhog (approved):
Hacking a little locally to fix the data returned from platform-api makes it look better. So these changes LGTM

PR #119, opened 2021-09-28
Add a storage class and persistent volume and docs

joelhoisko:
## Summary

Adds a persistent volume and storage class files for the local cluster development. They are called `managed-premium` so as to mock Azure's [`managed-premium`](https://docs.microsoft.com/en-us/azure/aks/concepts-storage#storage-classes) storage class that we use. This persistent volume is mounted on k3d cluster creation to `/tmp/k3dvolume`.

This is so that we can have persistence in our local cluster. Related to platform-api PR https://github.com/dolittle/platform-api/pull/36

### Added

- `managed-premium` storage class and persistent volume
- Docs on how to apply these during cluster bootstrapping

### DIFF
--- a/Environment/k3d/k8s/persistent-volume.yml
+++ b/Environment/k3d/k8s/persistent-volume.yml
@@ -1,0 +1,14 @@
+apiVersion: v1
+kind: PersistentVolume
+metadata:
+  name: managed-premium
+  labels:
+    type: local
+spec:
+  storageClassName: managed-premium
+  capacity:
+    storage: 10Gi
+  accessModes:
+    - ReadWriteOnce
+  hostPath:
+    path: "/tmp/k3dvolume"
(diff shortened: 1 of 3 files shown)


Environment/k3d/k8s/persistent-volume.yml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v1
+kind: PersistentVolume
+metadata:
+  name: managed-premium
+  labels:
+    type: local
+spec:
+  storageClassName: managed-premium
+  capacity:
+    storage: 10Gi
```
joelhoisko, line 10: See https://github.com/dolittle/platform-api/pull/36/files#r717593678 about the size

joelhoisko:
@freshteapot I assume this one is also then fine to merge in

PR #133, opened 2021-11-04
Buttons to download credentials

freshteapot:
## Summary

- Able  to download service account credentials, making it easier to setup azure pipeline.
- Able to download container registery, making it easier to setup azure pipeline.

### DIFF
--- a/Source/SelfService/Web/api/cicd.ts
+++ b/Source/SelfService/Web/api/cicd.ts
@@ -1,0 +1,38 @@
+// Copyright (c) Dolittle. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+import { getServerUrlPrefix } from './api';
+
+// @throws {Error}
+export async function getKubernetesServiceAccount(applicationId: string): Promise<any> {
+    const url = `${getServerUrlPrefix()}/application/${applicationId}/cicd/credentials/service-account`;
+
+    const response = await fetch(url, {
+        method: 'GET',
+        mode: 'cors',
+    });
+
+    if (response.status !== 200) {
+        console.error(response);
+        throw Error('Failed to service account');
+    }
+
+    return await response.json() as any;
+}
+
+// @throws {Error}
+export async function getContainerRegistry(applicationId: string): Promise<any> {
+    const url = `${getServerUrlPrefix()}/application/${applicationId}/cicd/credentials/container-registry`;
+
+    const response = await fetch(url, {
+        method: 'GET',
+        mode: 'cors',
+    });
+
+    if (response.status !== 200) {
+        console.error(response);
+        throw Error('Failed to service account');
+    }
+
+    return await response.json() as any;
+}
+
--- a/Source/SelfService/Web/api/cicd.ts
+++ b/Source/SelfService/Web/api/cicd.ts
@@ -25,7 +25,7 @@ export async function getAzureDevopsKubernetesServiceAccount(applicationId: stri
 
     if (response.status !== 200) {
         console.error(response);
-        throw Error('Failed to service account');
+        throw Error('Failed to get azure-devops service account');
     }
 
     return await response.json() as KubernetesServiceAccountCredentials;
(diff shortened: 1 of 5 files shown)


Source/SelfService/Web/documentation/container.tsx
```diff
@@ -81,6 +82,15 @@ export const DocumentationContainerScreen: React.FunctionComponent<Props> = (pro
                                     Verify access to kubernetes
                                 </a>
                             </li>
+                            <li>
+                                <a href="#" onClick={(event) => {
+                                    event.preventDefault();
+                                    const href = `/documentation/application/${applicationId}/${environment}/ci-cd/-azure-pipelines`;
```
joelhoisko, line 88:
```suggestion
                                    const href = `/documentation/application/${applicationId}/${environment}/ci-cd/azure-pipelines`;
```

Was the `-` prefix intentional? Looks a bit odd to me.

Source/SelfService/Web/documentation/container.tsx
```diff
@@ -107,6 +117,17 @@ export const DocumentationContainerScreen: React.FunctionComponent<Props> = (pro
                         <VerifyKubernetesAccess {...docProps} />
                     </Route>
 
+                    <Route exact path="/documentation/application/:applicationId/:environment/ci-cd/-azure-pipelines">
```
joelhoisko, line 120:
```suggestion
                    <Route exact path="/documentation/application/:applicationId/:environment/ci-cd/azure-pipelines">
```

Same here with the `-` prefix.

Source/SelfService/Web/documentation/cicd/setupAzurePipelines.tsx
```diff
@@ -0,0 +1,58 @@
...
+                const data = await getKubernetesServiceAccount(applicationID);
+                await navigator.clipboard.writeText(data);
+                enqueueSnackbar('Kubernetes service account copied to clipboard.');
+            } catch {
+                enqueueSnackbar('Failed to get data.', { variant: 'error' });
+            }
+        }}
+    >
+        Copy Credentials to clipboard
+    </DownloadButton>;
```
joelhoisko, line 30: The `<DownloadButton>`'s look a bit weird as you're just copying credentials, but I also know that mui v4 icons don't have the nice copy icon yet. In the future we should make a `<CopyButton>` for this. Just pointing this out as a paper trail https://app.asana.com/0/1200181647276434/1201312046888451

joelhoisko (changes requested):
Thinking about this, the information here is not specific to Azure Pipelines at all so maybe we could just call it something like "CI/CD Credentials" or something.

On the other hand, content is king and when we get something a bit nicer we can do some renaming.

Otherwise LGTM, just a small point about the URL.

Source/SelfService/Web/api/cicd.ts
```diff
@@ -0,0 +1,38 @@
...
+        method: 'GET',
+        mode: 'cors',
+    });
+
+    if (response.status !== 200) {
+        console.error(response);
+        throw Error('Failed to service account');
+    }
+
+    return await response.json() as any;
```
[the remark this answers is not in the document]
freshteapot: I have changed this to use a type not any, for both (you see it in my latest commit)

At least on my vscode, my opting to await is throwing errors when I remove the casting.
So I have kept it.

Source/SelfService/Web/documentation/container.tsx
freshteapot: Fixed, good spot.

Source/SelfService/Web/documentation/cicd/setupAzurePipelines.tsx
freshteapot: Changed to ButtonText.

Source/SelfService/Web/documentation/container.tsx
freshteapot: Fixed, good spot.

Source/SelfService/Web/api/cicd.ts
```diff
@@ -0,0 +1,38 @@
...
```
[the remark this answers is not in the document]
freshteapot: I have changed this to use a type not any, for both (you see it in my latest commit)

At least on my vscode, my opting to await is throwing errors when I remove the casting.
So I have kept it.

joelhoisko (approved):
👍

joelhoisko:
Also I already marked our conversation about the `as any` stuff as resolved, but here's a nice example explaining the some of the difference between the return type annotation vs using the `as type` stuff. https://stackoverflow.com/a/61386586/5806412

PR #136, opened 2021-11-15
Microservice without runtime

joelhoisko:
## Summary

Adds a new dropdown menu and uses it for selecting the Runtime image when creating a new base microservice. If `None` is selected for the Runtime image the microservice won't have a Runtime container created.

The new dropdown menu is styled according to this Figma design:
image

In action:
Peek 2021-11-15 15-20

Related to https://github.com/dolittle/platform-api/pull/61

Also adds `yarn.lock` into the repo.


joelhoisko:
# How to test

1. Reset your local k3d
2. Create a new base microservice, but select `None` in the Runtime image spot
3. Check that the created deployment only has one container in it:
```sh
kubectl -n application-11b6cf47-5d9f-438f-8116-0d9828654657 get deployments.apps -o wide
```

freshteapot (approved):
Works locally :)

PR #137, opened 2021-11-17
Add Grafana, Cortex & Prometheus to local k3d setup

joelhoisko:
## Summary

Adds our monitoring system to the k3d setup with docs.

Also enables cAdvisor metrics and adds kube-state-metrics setup.


freshteapot (approved):
- Followed the docs = success.
- The resources might be a little high, but I don't think we should worry right now.

Fabulous stuff!

PR #141, opened 2021-12-07
Fix to only show the environments microservices in the overview

joelhoisko:
## Summary

Fixes that only the microservices that belong to the environment are showed in the microservices overview screen.


freshteapot (approved):
Makes sense to me.

PR #148, opened 2022-01-19
Make the reset-local-k3d script more secure

joelhoisko:
## Summary

Makes the `reset-local-k3d` script check that the cluster creation actually worked and that it's in the correct context. This is so that we don't have any accidents in the future again when the context switching fails.

### DIFF
--- a/Environment/k3d/reset-local-k3d
+++ b/Environment/k3d/reset-local-k3d
@@ -27,13 +27,22 @@ k3d cluster create dolittle-dev \
 
 CREATE_EXIT_CODE=$?
 if [ $CREATE_EXIT_CODE -ne 0 ]; then
-    echo "Error creating cluster"
+    cat <<_EOF_
+
+Error switching to the k3d context
+You might need to clean up docker
+
+docker network rm k3d-dolittle-dev
+docker rm k3d-dolittle-dev-registry
+_EOF_
+
     exit $CREATE_EXIT_CODE
 fi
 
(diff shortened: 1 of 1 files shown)


Environment/k3d/reset-local-k3d
```diff
@@ -21,6 +25,17 @@ k3d cluster create dolittle-dev \
     --kubeconfig-switch-context \
     -v /tmp/k3dvolume:/tmp/k3dvolume
 
+CREATE_EXIT_CODE=$?
+if [ $CREATE_EXIT_CODE -ne 0 ]; then
+    echo "Error creating cluster"
```
freshteapot, line 30: We will need an exit here :)

Environment/k3d/reset-local-k3d
```diff
@@ -21,6 +25,17 @@ k3d cluster create dolittle-dev \
...
+fi
+
+kubectl config use-context k3d-dolittle-dev
+CONTEXT_EXIT_CODE=$?
+if [ $CONTEXT_EXIT_CODE -ne 0 ]; then
+    echo "Error switching to the k3d context"
```
freshteapot, line 36: We will need an exit here :)

freshteapot (changes requested):
Small fix :).

freshteapot:
@joelhoisko I added a commit, I hope the "bash" works on your machine, wanted to make it easier to get out of the problem, as I put myself in it, and then had to recall what we did yesterday over teams :).

PR #83, opened 2021-06-04
Download config files

freshteapot:
# Aim
- Able to download configmaps linked to the microservice

# Linked
- https://github.com/dolittle-entropy/platform-api/pull/5

# Reference
- https://app.asana.com/0/1200403658239260/1200315293318506/f

PR #150, opened 2022-02-02
Being specific about information from the backend

freshteapot:
## Summary

- Adding typing to the data coming via personalised-info endpoint.
- Able to override the DOLITTLE_CUSTOMER_ID via `make develop-backend`.


freshteapot (commented):
🚢 🇮🇹  works on dev with https://github.com/dolittle/platform-api/pull/83.

PR #151, opened 2022-02-14
Setting it to empty as it is not used on the view screen

freshteapot:
## Summary

Fixing a bug when trying to map ingress details to the microservice, ignoring it for now, as the view is not even using the extra.ingress info.


freshteapot (commented):
Confirmed on https://dev.dolittle.studio/selfservice/
- LMP
- Flokk
- Aksio

Whilst running the new version of the platform-api.

PR #159, opened 2022-03-03
Add new customer view

joelhoisko:
## Summary

Adds new customer view on `/admin/customer/{customerID}`, which can be navigated to by clicking on a customers name in `/admin/customers` page.

image

On this page you can toggle between enabling or disabling all of the customers environments. It currently supports only enabling them all (aka no disabled environments) or disabled them all (having a `*` wildcard entry in the array).

Also upgrades `@fluentui/react` package to `8.57.1`. This is because the frontend docker image had stopped mysteriously building earlier, upgrading the package fixed it.

## Reference
- [Related platform-api PR](https://github.com/dolittle/platform-api/pull/95)
- [Asana](https://app.asana.com/0/1201885260673030/1201894099742054)


freshteapot:
<3 the image @joelhoisko

Source/SelfService/Web/customer/view.tsx
```diff
@@ -0,0 +1,139 @@
...
+    return (
+        <>
+            <h1>Customer {customer.name}</h1>
+            <div className={classes.root}>
+                <h3>Customer Id: {customerId}</h3>
+                <h3>Build overwrite: {String(config.buildOverwrite)}</h3>
+                <h3>Can create applications: {String(config.canCreateApplication)}</h3>
+                <h3>Disabled environments: {config.disabledEnvironments}</h3>
+                <ButtonText
+                    onClick={() => {
```
freshteapot, line 119: I have now learnt I am very inconsitent, as all useEffects are promise with then and catch.
When really, I am all about using async and await.

Source/SelfService/Web/customer/viewAll.tsx
```diff
@@ -57,7 +53,9 @@ export const ViewAll: React.FunctionComponent<Props> = (props) => {
             <ul>
                 {customers.map((customer) => {
                     return <li key={`${customer.id}`}>
-                        <p>{customer.name} ({customer.id})</p>
+                        <a href={`/admin/customer/${customer.id}`}>
```
freshteapot, line 56: This doesn't work, it might do locally.
When running in dev or production it has the suffix "selfservice".

The current way I have this working is via history api.

```ts
const href = `/admin/customer/${customer.id}`;
history.push(href);
```

I am hopeful once we get @pavsaund into the code base, we will either have this cleared up, or someone will help me understand the insanity of react ;).

Source/SelfService/Web/customer/view.tsx
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    return (
+        <>
+            <h1>Customer {customer.name}</h1>
+            <div className={classes.root}>
+                <h3>Customer Id: {customerId}</h3>
+                <h3>Build overwrite: {String(config.buildOverwrite)}</h3>
+                <h3>Can create applications: {String(config.canCreateApplication)}</h3>
+                <h3>Disabled environments: {config.disabledEnvironments}</h3>
```
freshteapot, line 117: Can we list this, as it is an array of strings.

freshteapot (changes requested):
Nice progress @joelhoisko, I feel bad for the "/selfservice/" ugliness, but it needs to work in dev and production :)

Also the customers screen is really really ugly now. I think we either need to tame the a-tag or look to rely on javascript for basic functionality :( 
- Do we want to remove yarn.lock?

Source/SelfService/Web/customer/view.tsx
```diff
@@ -0,0 +1,139 @@
...
+                <ButtonText
+                    onClick={() => {
+                        config.disabledEnvironments = config.disabledEnvironments.length ? [] : ['*'];
+
+                        saveStudioConfig(customerId, config)
+                            .then(result => {
+                                if (!result) {
+                                    enqueueSnackbar('Failed saving the config', { variant: 'error' });
```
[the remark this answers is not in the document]
joelhoisko: Fixed

Source/SelfService/Web/customer/view.tsx
joelhoisko: Made it into an unsorted list.

Source/SelfService/Web/customer/view.tsx
```diff
@@ -0,0 +1,139 @@
...
+    const { enqueueSnackbar } = useSnackbar();
+
+    const { customerId } = useParams();
+    const [configLoaded, setConfigLoaded] = useState(false);
+    const [customerLoaded, setCustomerLoaded] = useState(false);
+    const [config, setConfig] = useState({} as Studio);
+    const [customer, setCustomer] = useState({} as Customer);
+    const [toggleConfig, setToggleConfig] = useState(false);
+
+    useEffect(() => {
```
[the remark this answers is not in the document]
joelhoisko: > I wonder if the customer endpoint should contain both data

I was thinking about that originally too, but as the customer data in the backend is compromised of multiple json files (`studio.json` and `customer.json`) I liked the simplicity of having an endpoint that just deals with one of them (as there already was an endpoint for getting just the `customer.json` data). I also liked that then I could have an endpoint that gives you just the data that you want (`studio.json`) and then accepts a POST request to save the studio config with the same structure.

What I'd ultimately like to do is to combine `studio.json` and `customer.json` but that option would be way more work so I didn't bother. I also didn't want to touch on the already existing code, but in hindsight I maybe should've just changed it and only had a single endpoint and fixed all of the other stuff too.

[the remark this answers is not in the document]
joelhoisko: But we need to make a decicision: to combine the endpoints now or later. The code already works so I'm in the camp of doing it later.

Source/SelfService/Web/customer/view.tsx
joelhoisko: I was thinking of using `async` and `await` inside the `useEffects`, but the way to do it looked [kinda ugly](https://devtrium.com/posts/async-functions-useeffect) so I didn't bother and we were already doing the `then()/catch()` stuff.

joelhoisko:
> Do we want to remove yarn.lock?
@freshteapot apparently [you should commit it.](https://stackoverflow.com/a/39992365/5806412)

Source/SelfService/Web/customer/view.tsx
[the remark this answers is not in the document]
freshteapot: But we need to make a decicision: to combine the endpoints now or later. The code already works so I'm in the camp of doing it later.

@joelhoisko, you decide, I support whatever you want todo (added a comment in teams, but this is my true feeling).

Source/SelfService/Web/customer/viewAll.tsx
freshteapot: I care :), its our little garden.
Nice work.

Source/SelfService/Web/customer/view.tsx
```diff
@@ -0,0 +1,150 @@
...
+                    onClick={() => {
+                        config.disabledEnvironments = config.disabledEnvironments.length ? [] : ['*'];
+
+                        saveStudioConfig(customerId, config)
+                            .then(result => {
+                                if (!result) {
+                                    enqueueSnackbar('Failed saving the config', { variant: 'error' });
+                                    return;
+                                }
+                                setToggleConfig(!toggleConfig);
```
freshteapot, line 138: A snackbar to say update or saved would be nice.

Source/SelfService/Web/customer/view.tsx
```diff
@@ -0,0 +1,150 @@
...
+                        Customer ID: {customerId}
+                    </li>
+                    <li>
+                        Build overwrite: {String(config.buildOverwrite)}
+                    </li>
+                    <li>
+                        Can create applications: {String(config.canCreateApplication)}
+                    </li>
+                    <li>
+                        Disabled environments: {config.disabledEnvironments}
```
freshteapot, line 125: I think when I suggested turning this into a list, there might have been a misunderstanding.
I am referring to config.disabledEnvironments.

Ie
["/fake-app-123/dev", "/fake-app-123/test"]

Vs

- /fake-app-123/dev
- /fake-app-123/test

freshteapot (approved):
Works for me!

I made some comments, but not show stoppers.

Source/SelfService/Web/customer/view.tsx
joelhoisko: Fixed it and also made it fancy by making it display `'none'` if there are no disabled environments instead of just nothingness.

PR #163, opened 2022-03-21
Fix direct links with environment not working

pavsaund:
## Summary
Opening direct links to microservices in an environment failed sometimes. The cause of this failure is related to which environment the user has last visited. If the user has last been in "Dev", and the link is pointing to "Prod", it would fail. This information is currently stored in local storage.

This PR introduces a higher-order component that reads the `environment` and `applicationId` parameters from the URL, and overwrites these in the global context, which also updates the value in local storage.

### Fixed

- Fix direct links to environments not working when pointing to another environment than the user last visited

### DIFF
--- a/Source/SelfService/Web/businessMoments/container.tsx
+++ b/Source/SelfService/Web/businessMoments/container.tsx
@@ -17,7 +17,7 @@ import '../application/applicationScreen.scss';
 import { useReadable } from 'use-svelte-store';
 import { businessmoments, load, isLoaded } from '../stores/businessmoment';
 import { microservices } from '../stores/microservice';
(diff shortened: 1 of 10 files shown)


pavsaund:
I manually approved codeclimate checks.

pavsaund:
Currently failing when clearing local storage and navigating directly to a microservice page.
image

freshteapot:
Future work https://app.asana.com/0/1200181647276434/1201993415401681/f

freshteapot (approved):
🚢 🇮🇹 

I have seen it work in https://dev.dolittle.studio/.

PR #170, opened 2022-04-04
Add suport for runtime version 8.0.0

pavsaund:
## Summary

Summary of the PR here. The GitHub release description is created from this comment so keep it nice and descriptive.

Remember to remove sections that you don't need or use.

### Added

- Describe the added features

### Changed

- Describe the outwards facing code change

### Fixed

- Describe the fix and the bug

### Removed

- Describe what was removed and why

### Security

- Describe the security issue and the fix

### Deprecated

- Describe the part of the code being deprecated and why

### DIFF
--- a/Source/SelfService/Web/api/api.ts
+++ b/Source/SelfService/Web/api/api.ts
@@ -108,8 +108,8 @@ export function getServerUrlPrefix(): string {
 
 export function getLatestRuntimeInfo(): LatestRuntimeInfo {
     return {
-        image: 'dolittle/runtime:7.8.1',
-        changelog: 'https://github.com/dolittle/Runtime/releases/tag/v7.8.1',
+        image: 'dolittle/runtime:8.0.0',
+        changelog: 'https://github.com/dolittle/Runtime/releases/tag/v8.0.0',
     };
 }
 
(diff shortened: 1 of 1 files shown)


Source/SelfService/Web/api/api.ts
```diff
@@ -108,17 +108,17 @@ export function getServerUrlPrefix(): string {
...
 }
 
 
 export function getRuntimes(): LatestRuntimeInfo[] {
     return [
         {
-            image: 'dolittle/runtime:7.8.1',
-            changelog: 'https://github.com/dolittle/Runtime/releases/tag/v7.8.1',
+            image: 'dolittle/runtime:8.0.0',
+            changelog: 'https://github.com/dolittle/Runtime/releases/tag/v8.0.0',
```
joelhoisko, line 121:
```suggestion
            getLatestRuntimeInfo(),
```

Just realized that we could just use the `getLatestRuntimeInfo()` method instead of having to keep this data up to date in 2 different places. Also my suggestion doesn't work as GitHub doesn't allow suggestions on deleted lines, but you get the idea. What do ya think?

pavsaund: I think I get it. Could you verify the change I made?

joelhoisko (approved):
:ship: :it:

PR #172, opened 2022-04-05
Making admin work in the local cluster

freshteapot:
## Summary

In the local cluster, the "local-dev" user now has access to view the admin screens.


freshteapot:
# Reference
- https://app.asana.com/0/1202023614957161/1202069684667496/f
- #171 (where it started before I added more and more)

PR #173, opened 2022-04-07
Making it possible to add customer tenant id to an environment

freshteapot:
## Summary

Add ability to manually specify multiple tenants for each environment when creating an application.


pavsaund:
Example image showing tenants added for each environment
image

jakhog (approved):

PR #185, opened 2022-05-04
Unique applications with their environments

freshteapot:
## Summary

- Fixed bug in customer screen within admin where view access would be repeated for each environment


freshteapot:
# Reference
- https://app.asana.com/0/1202121266838773/1202209556109798/f

PR #186, opened 2022-05-04
WIP: Spike to check how much work it is to view logs from Loki in Studio

jakhog:
Plan:
- [x] Verify that API is accessible from the frontend code and we have the data we need for a single microservice
- [x] Verify that custom searches can work
- [ ] Test log aggregation for an entire environment (might need some massaging of Loki response)

If we go ahead:
- [ ] Make local dev setup work better (currently 3 port-forwards and some header hacking involved)
- [ ] Make it look a little nicer


jakhog:
We have implemented this now in other PRs, so will close this.

PR #189, opened 2022-05-10
m3connector experience in Studio - version 1

freshteapot:
## Summary

Add m3connector link on the left to allow the user to interact with the m3 connector setup in the platform

## Added
- Able to get an overview of which environment has m3connector setup
- Able to view details about the m3 connector
- Able to trigger setup of the m3 connector (not talking to the backend)

Solves: https://app.asana.com/0/1202121266838773/1202253446321915/f


goeran:
A good first draft. It's only visible for users that uses the m3connector. Will merge.

PR #192, opened 2022-05-12
Feature-apply-theming

pavsaund:
## Summary

Introduce Dolittle Material Theme - Minimal visual changes, but a better inheritance of colors and ease of use as we go forward with future styles
- Dark color palette & typography based on Figma
- Introduce `<Typography>` component in place of headers across Studio

### DIFF
--- a/Source/SelfService/Web/App.tsx
+++ b/Source/SelfService/Web/App.tsx
@@ -17,7 +17,7 @@ import CssBaseline from '@mui/material/CssBaseline';
 import { RouteNotFound } from './components/notfound';
 import { Screen as AdminScreen } from './screens/adminScreen';
 
-import { createTheme, ThemeProvider, StyledEngineProvider } from '@mui/material/styles';
+import { ThemeProvider, StyledEngineProvider } from '@mui/material/styles';
 import { DieAndRestart } from './components/dieAndRestart';
 import { LayoutWithSidebar } from './layout/layoutWithSidebar';
 
@@ -30,6 +30,7 @@ import { ApplicationScreen } from './screens/applicationScreen';
 import { Box } from '@mui/material';
 import { ContainerRegistryScreen } from './screens/containerRegistryScreen';
 import { M3ConnectorScreen } from './screens/m3connectorScreen';
+import { themeDark } from './theme/theme';
 
 const typography = {
     fontFamily: 'Rubik',
--- a/Source/SelfService/Web/App.tsx
+++ b/Source/SelfService/Web/App.tsx
@@ -27,52 +27,11 @@ import { TransitionProps } from '@mui/material/transitions';
 import { ThemeScreen } from './screens/themeScreen';
 import '@fontsource/rubik';
 import { ApplicationScreen } from './screens/applicationScreen';
-import { Box } from '@mui/material';
+import { Box, createTheme } from '@mui/material';
 import { ContainerRegistryScreen } from './screens/containerRegistryScreen';
 import { M3ConnectorScreen } from './screens/m3connectorScreen';
 import { themeDark } from './theme/theme';
 
-const typography = {
-    fontFamily: 'Rubik',
-    fontStyle: 'normal',
-    htmlFontSize: 14,
-    fontSize: 14,
-    body1: {
-        fontFamily: 'Rubik',
-        color: '#E9EAEC',
-        fontSize: '14px',
-        letterSpacing: '-0.01em',
-        lineHeight: '16px'
-    }
-};
-
-
-const themeDark = createTheme({
-    typography,
-    palette: {
-        primary: {
-            main: '#8C9AF8',
-            light: '#b3bbfb',
-            dark: '#6678f6',
-            contrastText: '#191A21',
-        },
-        secondary: {
-            main: '#76E8DB',
-            light: '#A3EFE7',
-            dark: '#48E0CF',
-            contrastText: '#191A21',
-        },
-        background: {
-            default: '#0F1014',
-            paper: '#191A21',
-        },
-        text: {
-            disabled: '#93959f',
-            primary: '#fafafa',
-            secondary: '#cecfd0',
-        },
-    },
-});
 
 const snackbarStyles = {
     '& .SnackbarContent-root.': {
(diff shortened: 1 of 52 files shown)


Source/SelfService/Web/App.tsx
```diff
@@ -27,51 +27,11 @@ import { TransitionProps } from '@mui/material/transitions';
 import { ThemeScreen } from './screens/themeScreen';
 import '@fontsource/rubik';
 import { ApplicationScreen } from './screens/applicationScreen';
-import { Box } from '@mui/material';
+import { Box, createTheme } from '@mui/material';
 import { ContainerRegistryScreen } from './screens/containerRegistryScreen';
 import { M3ConnectorScreen } from './screens/m3connectorScreen';
+import { themeDark } from './theme/theme';
```
vstdolittle, line 33: Good move to start at the top of the stack for this. What i see already is lingo in code that can be used to communicate with others (than developers) touching the product. I believe, if you keep this in the back of your head when doing the design thinking moving forward,  some creative ideas/solutions could come out of it.

Source/SelfService/Web/admin/application/view.tsx
```diff
@@ -100,15 +101,15 @@ export const View: React.FunctionComponent<any> = (props) => {
 
     return (
         <>
-            <h1>Customer {customer.name}</h1>
+            <Typography variant='h1' my={2}>Customer {customer.name}</Typography>
```
vstdolittle, line 104: To give an example on how to extend the first comment in app.tsx: 
- designers and product owners dont care about in-depth knowledge on the workings of typography, but, would like to be able to refer to a component that does what they expect it to do. In this case it would be a h1 element with margin on the y axis 2. These are inputs they can reuse in Figma
-  They can create their own templates with their own naming. Since this line:
`<Typography variant='h1' my={2}>`
is reused, i wonder if it can be wrapped in a component which shares the name designers could reuse in Figma and behaves based on this. Not saying this should be done in this PR specifically, but its the opportunity i spot in this PR.

jakhog (approved):
:shipit:

Source/SelfService/Web/admin/application/view.tsx
vstdolittle: If ALL Typography header components use `my={2}` then i suggest to move it up to the theme level. That could be a an improvement done in this PR i find

vstdolittle: Since marginY is not a css property and actually consists of the following:
Screenshot 2022-05-16 at 11 27 26

You could solve that by setting the css properties in the screenshots to match the expected spacing 
further docs on spacing using themes: https://mui.com/system/spacing/

Source/SelfService/Web/theme/theme.ts
```diff
@@ -0,0 +1,111 @@
...
+    },
+};
+export const themeDark = createTheme({
+    palette: {
+        mode: 'dark',
+        primary: {
+            main: '#8C9AF8',
+            light: '#b3bbfb',
+            dark: '#6678f6',
+            contrastText: '#191A21',
```
vstdolittle, line 66: i've found it disappointing that MUI doesn't support storing of reused colours other than in the theme declaration where the same colours are obviously used in multiple places.
The quick solution is to copy paste the code. Mut perhaps this is another area where designers, product owners, and developers can speak the same language. And chose a name for a specific hexcolor code. This could then live in this same file  to beign with to make it easy to use as a map/js object of constant values.
Along these lines:
`const colors = {
DOLITTLE_BLACK_PRIMARY : '#191A21'
}`

This naming could then be referenced in Figma templates, asana tasks, etc...

vstdolittle: The goal here is not to create naming conventions that match Studio's domain, but rather from the Dolittle Design System POV.

Source/SelfService/Web/microservice/base/create.tsx
```diff
@@ -200,7 +200,7 @@ export const Create: React.FunctionComponent<Props> = (props) => {
                 </Grid>
 
                 <Grid item>
-                    <Typography component="p" sx={styles.data}>
+                    <Typography sx={styles.data}>
```
vstdolittle, line 203: i cant find any code that sets default typography component to p. Is this resolved by the library itself on runtime? curious since it looks cleaner this way

vstdolittle:
General question around the direction of the design system being developed. To what extent do we take responsiveness into consideration from the first steps being made? Do we start by saying we will support all screen sizes from mobile -> tablet -> laptop -> monitor -> big screen? or will this be iterated on at a later stage. Many things unknown at this point i understand, however, you are probably well aware of this, but responsive UI is not something one enjoys refactoring at a later stage. So maybe conversations can be held with the right people to the reach and limits of this design system? 

just a thought :)

Source/SelfService/Web/admin/application/view.tsx
pavsaund: I could have been a bit more clear on the "why" of using the `my={2}`- approach. Basically my idea is that this is a drop-in replacement for all headers currently in the app to lean on Typography element to pick up on the theme's font sizing. Without the Typography element, the styles were being pulled from the browser style - ie, with margin top and bottom for each header tag. By default MUI's Typography has no spacing, so I made a judgement call of 3 spacing elements not to break all the pages where headers were used.

I didn't want to spend time adjusting and amending each implementation, as some pages don't work, others are semi-working, and most are not styled correcyl. Nor am I interested in introducing a global `Title` - element at this time, because I don't consider the current stylings to be the source of truth from a component perspective. Moving forward when implementing more from Figma, I foresee these `Typography` elements to be replaced, and inline-styling them makes that easy to do on a case by case basis.

What I did do was verify one thing, which was cool. even though there is quite a bit of duplication in code, this translates to only 3 CSS rules generated across the app - which is an optimization that is done during the build ( I am unclear where it is happening though)

Source/SelfService/Web/theme/theme.ts
pavsaund: Oh, I really like this - the thought had come to mind as I was implementing this, but I must have lost it along the way :)

What I really like though is the additional bonus of naming the colors, so that they can be referenced in across disciplines.

I've added a follow-up task for this: https://app.asana.com/0/1201325052247030/1202282345093714/f

PR #193, opened 2022-05-20
start stop k3d scripts

vstdolittle:
## Summary

Make it easier for developers to work for short periods on studio. By having a start/stop script one can save the state in studio under development and get back to it at a later stage. reset-local-k3d would clear all volumes.

### Added

- start-local-k3d
- stop-local-k3d

### Changed

- Added documentation for the new scripts

### DIFF
--- a/Documentation/k3d.md
+++ b/Documentation/k3d.md
@@ -23,6 +23,10 @@ Run the reset-local-k3d script, that automatically deletes the existing k3d clus
 ```sh
 ./Environment/k3d/start-local-k3d
 ```
+## Stop the cluster
+```sh
+./Environment/k3d/stop-local-k3d
+```
 
 # For manual steps, continue following the guide.
 
--- a/Documentation/k3d.md
+++ b/Documentation/k3d.md
@@ -18,7 +18,13 @@ Run the reset-local-k3d script, that automatically deletes the existing k3d clus
 ./Environment/k3d/reset-local-k3d
 ```
 
-For manual steps, continue following the guide.
+## Start the cluster
+> NOTE: This requires the cluster to be setup in order to work. Follow either the manual steps or run reset-local-k3d
+```sh
+./Environment/k3d/start-local-k3d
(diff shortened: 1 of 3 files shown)


Documentation/k3d.md
```diff
@@ -18,7 +18,17 @@ Run the reset-local-k3d script, that automatically deletes the existing k3d clus
 ./Environment/k3d/reset-local-k3d
 ```
 
-For manual steps, continue following the guide.
+## Start the cluster
+> NOTE: This requires the cluster to be setup in order to work. Follow either the manual steps or run reset-local-k3d
+```sh
+./Environment/k3d/start-local-k3d
```
joelhoisko, line 24:
```suggestion
k3d cluster start dolittle-dev
```

We actually only need one line in the script, the `k3d` command works globally so we don't have to `cd` anywhere. So I don't think that you even need the separate script files, just documenting the `k3d cluster start/stop dolittle-dev` is enough.

Documentation/k3d.md
```diff
@@ -18,7 +18,17 @@ Run the reset-local-k3d script, that automatically deletes the existing k3d clus
...
+```
+## Stop the cluster
+```sh
+./Environment/k3d/stop-local-k3d
```
joelhoisko, line 28:
```suggestion
k3d cluster stop dolittle-dev
```

Same thing here, it's enough to just document the command.

joelhoisko (changes requested):
Thanks for the PR, docs and helper scripts are always nice. In this case I don't think that we actually need the scripts in their current form as they are just oneliners after all. If the scripts did a bit more (like an intelligent start/stop based on the state of the cluster) it would make sense to have them, but otherwise they act just as an extra layer of abstraction.

jakhog:
As this PR has been laying dead here for a while now, I've taken the liberty of applying Joels suggestion - and I will get them in. Thanks Victor!

PR #198, opened 2022-06-03
Only show microservices in current env in filter

pavsaund:
## Summary
 - Fixes so that the microservice filter dropdown only shows microservices for the current environment


jakhog (approved):
:shipit:

PR #203, opened 2022-06-14
SelfService Backend mock for logs

jakhog:
## Summary

Adds a new Node ExpressJS server for running a mocked backend to see if it is useful for SelfService web development.


jakhog:
Oops, when looking at the commits - I see that I also made a small improvement to the `useLogsFromLast` code. Well well, that will get out together with the next release. I don't think anyone will notice a difference anyways.

PR #204, opened 2022-06-14
Show shimmering lines while fetching log lines

jakhog:
## Summary

Adds shimmering LogLines in a LogPanel that is currently loading more logs. Also fixed the "No logs found" message so that it does not appear while loading logs.

Note: A slightly annoying behaviour is - if you toggle the Timestamp while shimmering lines are shown, the animation is not in sync with the text. But this is kind of edge case, as the loading doesn't really appear for long.

image

### Changed

- The `useLogsFromLast` no longer retains the last loaded logs while loading more, this seems more in line with the designed UX with these new shimmering lines.

### Fixed

- A tiny bug in `useLogFromRange` caused it to not update the state to `loading: true` while fetching logs.

### DIFF
--- a/Environment/mock/monitoring/logs/generated.js
+++ b/Environment/mock/monitoring/logs/generated.js
@@ -59,10 +59,6 @@ generateRandomLogsEveryOtherSecond(generated);
 
 
 exports.queryGeneratedLogs = (query, from, to, limit, direction) => {
-    if (direction === 'forward') {
-        throw new Error('Direction forward is not implemented');
-    }
-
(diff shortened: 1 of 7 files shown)


jakhog:
I see that I have not matched the Figma design properly. The sketch shows 1x3 lines of shimmer, I did 2x2. Easy fix however we want to do it, I kind of like having more than one shimmering timestamp. But I don't mind all that much.

pavsaund (approved):
So much shimmer! :shipit:

PR #206, opened 2022-06-16
Keep Logs screen filter state in URL so it can be shared (and reloaded), and keep it when switching environments.

jakhog:
## Summary

This PR does two things:
1) It stores the state of the `LogsScreen` in the URL search querystring, this means that it can be shared with others and you can reload the page without resetting the filters.
2) The `ApplicationsChanger` component reuses whatever URL search querystring for the current URL when switching environments. Together with the first change, this means that switching environments/apps while looking at logs will keep the filters after switching environments.

While working on it, I found a little bit of a strange setup with multiple routers that made the querystring changes not propagate properly - and I don't know if that is intended usage. But I think we can leave this for later. Also 2) is a little bit hacky, but I don't think we should bother with doing it more properly (passing in props) before we have had a chance to revisit the structure and state management of the app as a whole.


Source/SelfService/Web/screens/logsScreen.tsx
```diff
@@ -35,10 +36,59 @@ import {
 import { TopNavBar } from '../components/topNavBar';
 import { HttpResponseApplication, getApplications, getApplication, HttpResponseApplications } from '../api/application';
 
-import { LogFilterMicroservice, LogFilterObject, LogFilterPanel } from '../logging/logFilter/logFilterPanel';
+import { LogFilterMicroservice, LogFilterObject, LogFilterDateRange, LogFilterPanel } from '../logging/logFilter/logFilterPanel';
 import { LogPanelRelative } from '../logging/logPanelRelative';
 import { LogPanelAbsolute } from '../logging/logPanelAbsolute';
 
+const useFilters = (initialFilters: LogFilterObject, availableMicroservices: LogFilterMicroservice[]): [LogFilterObject, React.Dispatch<React.SetStateAction<LogFilterObject>>] => {
```
pavsaund, line 43: The `useFilters` hook can be moved to the logging feature folder. 
Perhaps renaming it to `useLogFilters` also makes sense?

Source/SelfService/Web/package.json
```diff
@@ -32,13 +32,16 @@
...
+        "query-string": "^7.1.1",
         "react-markdown": "^7.1.0",
         "remark-gfm": "^3.0.1",
         "rxjs": "^7.5.5",
         "url-loader": "^4.1.1",
+        "use-query-params": "^1.2.3",
         "use-react-router-breadcrumbs": "2.0.2"
     },
     "devDependencies": {
+        "@types/react-router-dom": "^5.3.3",
```
pavsaund, line 44: This change caused typescript to fall over with errors on my machine for places we are using `useParams()`. Could you spend a few minutes to update the typings for this?

image

pavsaund (approved):
I have a few suggestions on the review that are easy to fix.
I have tested this locally and it works just fine.

pavsaund:
@jakhog, before merging this in: there is a blocker in main because of https://github.com/dolittle/Studio/pull/208, which has introduced a bug / slowness on the page which should be fixed.

Source/SelfService/Web/screens/logsScreen.tsx
jakhog: Done!

jakhog:
Nice, thanks @pavsaund 😄 I've fixed the merge conflicts, TS build issues and implemented your suggestions. Also found a small bug in the `ApplicationsChanger` that created strange URLs and made me click the changer twice to get it to work.

I'm still not proud of this `useLogFilters` hook, it's quite big and does a lot of default things. But I think it might be okay for now - as we've discussed briefly, this might be something worth building in a reusable way together with routing. There are too many libraries out there to find a good one.

Lastly, which I'll leave up to you to decide - is __whether we want it to replace or push in the history__, on this line: https://github.com/dolittle/Studio/blob/8ea4ff12f88b77f2a3a224c9b75fdd4bdab92064/Source/SelfService/Web/logging/logFilter/useLogFilters.ts#L52

I think it would be annoying UX to have to click back a lot of times when you've played with the filters. But second opinions are nice 🙂 If you agree, it should be ready to merge.

pavsaund:
:shipit:

PR #209, opened 2022-06-21
Update studio header information

N00bG1rl:
## Summary

The appearance of the page has been changed in the browser tab.

Screenshot 2022-06-21 at 14 09 45

### Added

Added favicon in two formats: in most cases svg format, but also in favicon.ico format for edge cases.

### Changed

 Changed page tiltle tag from 'Studio' to 'Dolittle Studio' for better UX.


Source/SelfService/Web/index.ejs
```diff
@@ -7,6 +7,7 @@
     <title>
         <%- htmlWebpackPlugin.options.metadata.title %>
     </title>
+    <link rel="shortcut icon" href="{%=o.htmlWebpackPlugin.files.favicon%}" >
```
pavsaund, line 10: It looks like the HtmlWebpackPlugin had updated behaviour, and this is not needed anymore. It also has the wrong syntax, but that isn't relevant.

have a look at the comments on this SO answer: https://stackoverflow.com/questions/35239733/webpack-html-webpack-plugin-load-favicons-in-template#comment86983431_49595058

pavsaund (changes requested):
This is great! I've learned that the HtmlWebpackPlugin automatically generates the necassary meta tag for favicon.svg, and it falls back to favicon.ico if it isn't supported.

I don't see any meta tags being generated for devices, as you mentioned - perhaps this is a setting that needs to be changed?

I've also added a few comments that neeed to be looked at :)

pavsaund (approved):
It makes sense to default to png, as that is the most widely supported standard at this time.

pavsaund:
Feel free to merge this change @N00bG1rl!
:shipit:

PR #210, opened 2022-06-22
Removed code that includes new static microservice

N00bG1rl:
## Summary

Removed code that involves 'create new static site microservice' as it is no longer in use.

### Removed

Removed Card and options in code to select 'kind=static-site' .


pavsaund:
Great! Getting rid of some cruff as we make room for new improvements.

:shipit: 👍

pavsaund (approved):

PR #211, opened 2022-06-23
Design system components

soumikgm:
## Summary

This PR brings in the Design System Component library from [Dolittle Entropy ](https://github.com/dolittle-entropy/design-system) to a standalone project in Studio.

### Added

- Storybook that allows us to interact with different states of a component (stories) without having to run code locally.
- A first component called `Button` that leverages MUI at its core with a wrapper on top to suit our needs.
- A folder structure that follows the atomic design
- Theming to view our atoms/molecules/organisms/pages like the way we intend to use them


pavsaund (approved):

pavsaund:
We need to create a follow-up task so that this library can be consumed by Studio - I'll do that. Let's ship this and see if the deployment pipeline kicks in on main

PR #213, opened 2022-06-28
Make search case-insensitive and update placeholder text style

pavsaund:
## Summary

- Search filter placeholder text style changed to italic
- Updated search to be case insensitive

image

### DIFF
--- a/Source/SelfService/Web/logging/logFilter/logFilterToLabelsAndPipeline.ts
+++ b/Source/SelfService/Web/logging/logFilter/logFilterToLabelsAndPipeline.ts
@@ -22,7 +22,10 @@ export const logFilterToLabelsAndPipeline = (applicationId: string, environment:
                 : undefined,
     };
 
-    const pipeline = filters.searchTerms.map((term) => `|~"(?i)${term}"`);
+    const pipeline = filters.searchTerms
+        .map((term) => term.replace(/[.*+?^${}()|\\"]/g, '\\$&'))
+        .map((term) => term.replace(/[[\]]/g, '\\\\$&'))
+        .map((term) => `|~"(?i)${term}"`);
 
     return [labels, pipeline];
 };
--- a/Source/SelfService/Web/logging/logFilter/logFilterToLabelsAndPipeline.ts
+++ b/Source/SelfService/Web/logging/logFilter/logFilterToLabelsAndPipeline.ts
@@ -22,7 +22,7 @@ export const logFilterToLabelsAndPipeline = (applicationId: string, environment:
                 : undefined,
     };
 
-    const pipeline = filters.searchTerms.map((term) => `|= "${term}"`);
+    const pipeline = filters.searchTerms.map((term) => `|~ "${term}"`);
 
     return [labels, pipeline];
 };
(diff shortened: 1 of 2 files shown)


pavsaund:
@jakhog: please spend a few minutes looking at this. I think it should be straightforward to verify. The PR branch is deployed to "dev"

jakhog (changes requested):
So, the main point of making it case insensitive works 😄 

But sorry, I forgot one thing. Now that we are actually searching with RegExp, there are some new quirks for special characters. For example, if you now put "[5]" in the term, this is interpreted as any character in the group [5], so it will in practice ignore the square braces.

To fix this, we need to run the search terms through an escaper first. I think we can use this suggestion from here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#escaping.

It would be a good idea to also escape `"` characters (this was an existing bug). So I think adding another map to the searchTerms with something like:
```js
.map((term) => term.replace(/[.*+?^${}()|[\]\\"]/g, '\\$&'))
```
would do the trick.

Source/SelfService/Web/logging/logFilter/logFilterToLabelsAndPipeline.ts
```diff
@@ -22,7 +22,7 @@ export const logFilterToLabelsAndPipeline = (applicationId: string, environment:
                 : undefined,
     };
 
-    const pipeline = filters.searchTerms.map((term) => `|= "${term}"`);
+    const pipeline = filters.searchTerms.map((term) => `|~"(?i)${term}"`);
```
jakhog, line 25:
```suggestion
    const pipeline = filters.searchTerms
        .map((term) => term.replace(/[.*+?^${}()|[\]\\"]/g, '\\$&'))
        .map((term) => `|~"(?i)${term}"`);
```

pavsaund: Aha, so we're basicaly sending RegExp to loki to be parsed there. So with this escaping it would abstract that away from the user using the search field, so they could only search for plain text?

Source/SelfService/Web/logging/logFilter/logFilterToLabelsAndPipeline.ts
```diff
@@ -22,7 +22,10 @@ export const logFilterToLabelsAndPipeline = (applicationId: string, environment:
                 : undefined,
     };
 
-    const pipeline = filters.searchTerms.map((term) => `|~"(?i)${term}"`);
+    const pipeline = filters.searchTerms
+        .map((term) => term.replace(/[.*+?^${}()|\\"]/g, '\\$&'))
+        .map((term) => term.replace(/[[\]]/g, '\\\\$&'))
```
pavsaund, line 27: @jakhog I ended up having to double eescape the square brackets.

Source/SelfService/Web/logging/logFilter/logFilterToLabelsAndPipeline.ts
pavsaund: After testing this out a bit locally, I ended up having to double-escape the special characters.

pavsaund: This also matches what is sent when testing this regexp in grafana: `|~"(?i)\\?"`. Generates the same request to the server :)

jakhog (approved):
Nice! :shipit:

PR #214, opened 2022-07-04
Fix studio side menu

soumikgm:
## Summary

This PR brings in a refreshed side menu to Studio. This enables users to see and navigate features in Studio better than before. Icons help you quickly recognize different pages, and you can easily see which page you are currently on by looking at the active item in the menu. Also, the menu had been divided into two sections to help the users distill the actions they can take from the menu.
image

### Added

- Mui Icons to the side menu items.
- Custom icons on the container registry menu item.
- New Dolittle Logo on the top.
- Hover and active status indicators for the menu items.
- A bottom item section for the side menu.

### Changed

- "Change customer" option has been moved from the top-right menu to the bottom of the side menu.
- Side menu is now rendered using MUI `<List>` component family instead of standard HTML `<ul>` and `<li>` components. 

### Fixed

- Fixed the styling to accommodate the changes mentioned above and to make it look like the mockups.


Source/SelfService/Web/layout/layout.scss
```diff
@@ -30,18 +31,22 @@ svg.logo {
 
   .sidebar ul {
     list-style-type: none;
+    padding: 0;
+    margin: 0;
```
pavsaund, line 35: I think it makes more sense to move this into the `sx`-attribute of the component, now that we have this possibility. The only caveat is if it causes a FOUC (Flash of unstyled content) by moving it in.

Source/SelfService/Web/layout/layoutWithSidebar.tsx
```diff
@@ -101,28 +62,81 @@ export const LayoutWithSidebar: React.FunctionComponent<Props> = (props) => {
 
 export const getDefaultMenuWithItems = (
     history: History<LocationState>,
-    items: any[]
+    topItems: any[],
+    bottomItems: any[]
```
pavsaund, line 66: Would you consider renaming `topItems` => `navigationItems`, and `bottomItems` => `actionItems`, or something similar. 

The reason is I usually try to lean on the specific semantics of the properties as much as possible, and not focus on the pure visual aspect, like top/bottom. Now, this isn't something that's always avoidable, and for purely visual components then it makes sense to use these visual terms.

Thoughts?

Source/SelfService/Web/layout/layoutWithSidebar.tsx
```diff
@@ -101,28 +62,81 @@ export const LayoutWithSidebar: React.FunctionComponent<Props> = (props) => {
...
         <>
-            <ul>
-                {items.map((link) => {
+            <List>
+                {topItems.map((link) => {
+                    return (
+                        <ListItemButton
+                            disableGutters
+                            key={link.name}
+                            selected={window.location.href.includes(link.href)}
```
pavsaund, line 76: This is such a nice, sweet way to resolve the selected state! :)
There are some potential pitfalls if multiple url's have the same paths, which would lead to multiple items being selected, but I don't see this happening any time soon. Let's use this and improve 👍

pavsaund (approved):
This is so nice to see come to life! There are a few things I did see when reviewing this case, which I thought I'd bring up. I've added some comments in the code, and there are some comments from a purely visual aspect which would be worth difgging in to.

Here are the visual issues I registered when comparing the implementation with Figma, and testing on my machine. I also had a chat with Liz to go over these aspects:

* There should be more width on the menu in expanded state (260px in Figma)
* The background has the correct color, but wrong elevation - you can use Paper with an elevation of 4 (you can see this value in Figma)
* The Icons are a slightly different color - look at Figma to see what is being used there. Unsure if it's available in the palette or if it needs to be introduced
* The Menu scrolls with the content (ex: Logs) - and this looks very strange, especially now that there is a sticky footer. It should be fixed, and the main content should have the scroll.

In the name of iteration, I suggest you dig into the low-hanging fruit and fix those as part of this PR. For more involved changes, i suggest introducing "improvement" tasks, ex: the scroll one may be a bigger change.

Source/SelfService/Web/components/topRightMenu.tsx
```diff
@@ -45,11 +45,11 @@ export const TopRightMenu: React.FunctionComponent<Props> = (props) => {
             <NotificationsIcon sx={styles.icon} />
         </IconButton>
 
-        <AccountSettingsMenu child={
+        {/* <AccountSettingsMenu child={
```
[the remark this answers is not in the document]
soumikgm: Remove this component

Source/SelfService/Web/layout/layoutWithSidebar.tsx
soumikgm: yayyy

.vscode/settings.json
```diff
@@ -1,61 +1,62 @@
...
+    "fluentui",
+    "MUUID",
+    "mvvm",
+    "Stubbable",
+    "tsoa",
+    "tsyringe",
+    "typegoose",
+    "upsert"
+  ],
+  "editor.defaultFormatter": "esbenp.prettier-vscode"
```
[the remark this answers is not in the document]
soumikgm: Delete this file as we will have another task dealing with formatting across the whole codebase.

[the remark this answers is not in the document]
soumikgm: Done

Source/SelfService/Web/components/topRightMenu.tsx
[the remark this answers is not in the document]
soumikgm: Done!

Source/SelfService/Web/layout/layoutWithSidebar.tsx
soumikgm: Done!

pavsaund:
Nice! thanks for the improvements and discussions @soumikgm & @N00bG1rl ! Let's get this 🚢  :shipit:

pavsaund (approved):

PR #217, opened 2022-07-07
Fix change customer link

soumikgm:
## Summary

Fixes the issue of changing the customer button not working properly by forcing a reload of the webpage instead of using the router.


soumikgm:
Merging this in

PR #218, opened 2022-07-07
Login flow improvements

N00bG1rl:
## Summary

Improved login flow. This is the first page user sees after login. User can create new application, which includes the application name, contact name, contact email and selecting which environments they would like their application to have.

Screenshot 2022-07-19 at 16 15 52

### Added

1. Created layout without left sidebar.
2. Created right sided layout.
3. Basic form validation.

### Changed

-Removed stepper from 'Create new application' and reaplaced it with form.
-Refactored 'create' component into multiple smaller ones.

### Fixed

-General styling and updated theme.

### Removed

-Unnecessary code.


Source/SelfService/Web/application/create.tsx
```diff
@@ -3,140 +3,154 @@
...
+    supportLink: {
+        color: themeDark.palette.primary.main,
+        textDecoration: 'underline',
+        cursor: 'pointer',
+    }
+};
+
+interface ErrorStates {
+    applicationNameError: {
+        message: String,
```
soumikgm, line 80: I think the type here should be `string` and not `String`. `String` with a capital `S` is a javascript string type, which can be used to create new strings like this `const someString = new String("This is a string")`. `string` with a small `s` is a typescript string type. You can read more about this [here](https://stackoverflow.com/questions/14727044/what-is-the-difference-between-types-string-and-string).

Source/SelfService/Web/application/create.tsx
```diff
@@ -3,140 +3,154 @@
...
-    const steps = [
-        'Provide a name',
-        'Provide contact info',
-        'Setup Environments'
-    ];
-
 
-    const [activeStep, setActiveStep] = React.useState(0);
+    const newApplicationId = Guid.create().toString();
     const [application, setApplication] = React.useState({
```
soumikgm, line 137: instead of writing `React.useState` multiple times, perhaps we could do `import { React, useState } from "react"` in line 4?

Source/SelfService/Web/application/create.tsx
```diff
@@ -317,64 +198,183 @@ export const Create: React.FunctionComponent<Props> = (props) => {
...
             }
+
             return;
         }
 
-        setActiveStep((prevActiveStep) => prevActiveStep + 1);
+        return;
+    };
+
+    const validateForm = () => {
```
soumikgm, line 215: This file is getting big. Do you think we should move this validation function into its own file and then import it here? Perhaps create a folder called helperFunctions and put it inside there? This will require rewriting it a bit to pass all the necessary arguments to it when it is being called. Thoughts?

Source/SelfService/Web/application/create.tsx
```diff
@@ -317,64 +198,183 @@ export const Create: React.FunctionComponent<Props> = (props) => {
...
+            email.isError = false;
+        }
+
+        setFormError(errors);
+
+        return (!app.isError && !name.isError && !email.isError);
+    };
+
+    const handleCancel = () => {
+        // TODO: Add cancel confirmation modal.
```
soumikgm, line 254: why is this a TODO?

soumikgm (commented):
Great work @N00bG1rl ! I have made some comments on certain things where I have different thoughts, but otherwise, things look clear and fantastic.

Source/SelfService/Web/application/create.tsx
N00bG1rl: Really good idea :)

soumikgm (commented):
Great job again @N00bG1rl. Love how refactoring makes things very readable and pretty to look at. You have put a lot of work into this and it shows. I have a few comments on using styling and removing TODOs. So when you are done with those comments, tag me and I will approve it.

soumikgm (approved):
Great work. Everything looks good. :shipit:

PR #219, opened 2022-07-08
fix building of main branch by deleting design system

soumikgm:
## Summary

This PR fixes the failing build of Studio from the main branch. Running the code locally and on GitHub was failing due to some conflicts with yarn workspaces when the Design System folder was added to the Souce folder. This needs to be looked at in detail when we bring back the design system folder in main. For now, it has been deleted to fix the failing build. 

### Removed

- Design System folder from the code due to conflicts with yarn workspaces resulting in failing builds of the main branch locally and remotely.


soumikgm:
Merging this is because there is no CI/Docker workflow triggered when changes happen to the Design System folder. This is also the reason we never caught the main branch breaking when Design System was pushed into it. It never built the whole thing because there were no changes done to the Self Service or the Shared folder.

PR #225, opened 2022-08-08
Fixed menu item styles

N00bG1rl:
## Summary

Fixed menu dropdown styles when hovered and added padding.

Screenshot 2022-08-08 at 10 32 02

### Fixed

- Menu items hover and focused state style

### DIFF
--- a/Source/SelfService/Web/logging/logFilter/filterSelect.tsx
+++ b/Source/SelfService/Web/logging/logFilter/filterSelect.tsx
@@ -6,40 +6,34 @@ import React from 'react';
 import { Button, Select, SelectProps, Theme } from '@mui/material';
 
 const styles = {
-    wrapper: {
-        'height': '2.5rem',
-        '.MuiSelect-select:focus': {
-            backgroundColor: 'transparent'
-        },
-        '.MuiSelect-iconOpen': {
-            color: (theme: Theme) => theme.palette.primary.main
-        },
-        '& svg': {
-            right: '-0.3125rem'
-        },
-        '&:before, &:hover:not(.Mui-disabled):before, &:after': {
-            border: 'none',
-        },
-        ':hover': {
-            '& svg': {
-                color: (theme: Theme) => theme.palette.primary.main
-            }
-        },
+    'height': '2.5rem',
+    '.MuiSelect-select:focus': {
+        backgroundColor: 'transparent'
     },
-    inputField: {
-        textDecoration: 'none',
-        typography: 'button'
+    '.MuiSelect-iconOpen': {
+        color: (theme: Theme) => theme.palette.primary.main
+    },
+    '& svg': {
+        right: '-0.3125rem'
+    },
+    '&:before, &:hover:not(.Mui-disabled):before, &:after': {
+        border: 'none',
+    },
+    ':hover': {
+        '& svg': {
+            color: (theme: Theme) => theme.palette.primary.main
+        }
     }
 };
 
 export type FilterSelectProps<T> = SelectProps<T>;
 
 export const FilterSelect = <T,>(props: FilterSelectProps<T>) =>
-    <Button variant='text' disableRipple sx={styles.wrapper}>
+    <Button variant='text' disableRipple sx={styles}>
         <Select
             {...props}
             variant='standard'
             disableUnderline
-            sx={{ ...styles.inputField, ...props.sx }}
+            sx={{ typography: 'button', ...props.sx }}
         />
     </Button>;
--- a/Source/SelfService/Web/logging/logFilter/filterSelect.tsx
+++ b/Source/SelfService/Web/logging/logFilter/filterSelect.tsx
@@ -3,42 +3,42 @@
 
 import React from 'react';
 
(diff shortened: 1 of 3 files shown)


Source/SelfService/Web/logging/logFilter/filterSelect.tsx
```diff
@@ -2,8 +2,35 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 import React from 'react';
+
+import { themeDark } from '../../theme/theme';
 import { Select, SelectProps } from '@mui/material';
 
+const styles = {
+    'padding': '5px 14px',
+    'borderRadius': '4px',
```
pavsaund, line 11: Since these spacings effect the external paddings of the component, I think it would make sense to lean on MUI-spacing. The benefit of this is that if we ever change the fundamental look&feel of the theme, then the spacing will stay relative.

Docker has this in their MUI guidelines as well, which I think makes sense to lean on: https://docs.docker.com/desktop/extensions-sdk/design/mui-best-practices/

Source/SelfService/Web/logging/logFilter/filterSelect.tsx
```diff
@@ -2,8 +2,35 @@
...
+    'transition': 'all .3s',
```
pavsaund, line 12: I like that you are leaning on the transition to mimic what the default text button does. It would be good to lean on the same values as the button, though, and not have to maintain this specific transition. In our call earlier today, you found the transition duration and easing variables in MUI (https://mui.com/material-ui/customization/transitions/#durations), so I'm leaving this as a reference

```suggestion
    'transition': (theme: any) => theme.transitions.create(['all'], { duration: theme.transitions.duration.short }),
```

Source/SelfService/Web/logging/logFilter/filterSelect.tsx
```diff
@@ -2,8 +2,35 @@
...
+    '&:before, &:hover:not(.Mui-disabled):before, &:after': {
+        border: 'none',
+    },
+    ':hover': {
+        'background': 'rgba(140, 154, 248, 0.08)',
+        '& svg': {
+            color: themeDark.palette.primary.main
```
pavsaund, line 19: I'm glad you are using the theme color here. Instead of using it directly from the theme file, you should lean on the theme callback, so it always uses the correct color. This will also apply if we dynamically change themes (ex: Light/Dark)

I don't know the syntax off-hand, but something like:

```ts
    color: (theme) => theme.palette.primary.main
```

Source/SelfService/Web/logging/logFilter/filterSelect.tsx
```diff
@@ -2,8 +2,35 @@
...
+    },
+    'typography': 'button',
+    '& svg': {
+        right: '5px'
+    },
+    '.MuiSelect-select:focus': {
+        backgroundColor: 'transparent'
+    },
+    '.MuiSelect-iconOpen': {
+        color: themeDark.palette.primary.main
```
pavsaund, line 30: same comment as above :)

Source/SelfService/Web/logging/logFilter/filterSelect.tsx
```diff
@@ -12,9 +39,6 @@ export const FilterSelect = <T,>(props: FilterSelectProps<T>) =>
         variant='standard'
         sx={{
             ...props.sx,
-            '&:before, &:hover:not(.Mui-disabled):before, &:after': {
-                border: 'none',
-            },
-            'typography': 'button',
+            ...styles
```
pavsaund, line 42: Could you swap the order of these, so the `props.sx` is the last? This way `props.sx` property can be overridden by a developer consuming this component :)

Source/SelfService/Web/logging/logFilter/filterSelect.tsx
```diff
@@ -2,8 +2,35 @@
...
+    '& svg': {
+        right: '5px'
+    },
+    '.MuiSelect-select:focus': {
+        backgroundColor: 'transparent'
+    },
+    '.MuiSelect-iconOpen': {
+        color: themeDark.palette.primary.main
+    },
+};
```
pavsaund, line 32: I wish I knew what the type of those should be, so that you could get intellisense why accessing the `theme` component in the callbacks. If you figure it out, please do give it a type, if not don't worry, we can revisit it later. My first guess is `SxProps<Theme>` - but this didn't compile.

pavsaund (changes requested):

Source/SelfService/Web/logging/logFilter/filterSelect.tsx
pavsaund: the default spacing value (which is what we are using) is 1 = 8px, so 5px is 0.625, and 14px is 1.75. 

You can use the spacing variants, or a spacing util to specify multiple values ex: `padding: theme.spacing(0.625, 1.75)` or `px: 0.625, py: 1.75`.

More info and variants here: https://mui.com/material-ui/customization/spacing/#multiple-arity

PR #227, opened 2022-08-10
Bring DesignSystem back into action - and cleanup project and build setup

woksin:
## Summary

Brings back the DesignSystem and Storybook introduced by @soumikgm a while back. This required a bit of finagling with `tsconfig` and `package.json` files to bring it back into a building state. The final result is that we have a working Storybook that builds, and the components defined in there can be imported in the `SelfService/Web` project with e.g. `import { Button } from '@dolittle/design-system/atoms/Button'`. This works with WebPack and hot-reloading while editing (although a bit slow).

Further, to improve DX - I have reconfigured the build pipeline and checks in GitHub. Now, we build `DesignSystem`, `SelfService/Web` and `SelfService/Backend` on every push. This means that there is less of a chance of us inadvertently doing something that breaks one of these things without noticing in a PR. It also means that when merged to master - we will see a build result on the commit.


woksin:
The previously used versions of eslint did not work on my node version (17), updating them fixed the problem

woksin:
And in the process I disabled some rules for a couple of lines, since I don't know whether we should give regard to those rules

jakhog:
I have tested that this builds from a clean install (remove node_modules, yarn cache, dist and wwwroot folders) for both the DesignSystem and SelfService/Web. I also did a manual deploy of the branch to dev.dolittle.studio - and it starts up and runs normally. I will therefore merge and deploy this to dev for further testing before the main branch gets too far ahead.

PR #230, opened 2022-08-12
Fix data table

N00bG1rl:
## Summary

This PR will fix errors with microservices overview data table, when data is undefined.

Screenshot 2022-08-12 at 17 49 44

### Changed

- Refactored customUrlFieldSort function

### Fixed

- Microservices data table errors

### DIFF
--- a/Source/SelfService/Web/microservice/dataTable.tsx
+++ b/Source/SelfService/Web/microservice/dataTable.tsx
@@ -28,46 +28,50 @@ const styles = {
     }
 };
 
-export const sortByRuntimeVersion = (params: GridValueGetterParams) => {
-    const runtimeVersion = params.row.edit.extra.runtimeImage.replace(/dolittle\/runtime:/gi, '');
+const sortByRuntimeVersion = (params: GridValueGetterParams) => {
+    const runtimeVersion = params.row.edit?.extra?.runtimeImage?.replace(/dolittle\/runtime:/gi, '');
+    if (runtimeVersion === undefined) return 'N/A';
+
     return `${runtimeVersion || 'None'}`;
 };
 
-const tooltipCell = (params: GridRenderCellParams) => (
-    // TODO: Tooltip needs public urls. Map them into title.
-    <Tooltip title={``} arrow>
-        <span>{params.row.edit.extra.isPublic ? 'Available' : 'None'}</span>
-    </Tooltip>
-);
+const tooltipCell = (params: GridRenderCellParams) => {
+    const publicUrl = params.row.edit?.extra?.isPublic;
+    if (publicUrl === undefined) return 'N/A';
+
+    return (
+        // TODO: Tooltip needs public urls. Map them into title.
+        <Tooltip title={``} arrow>
+            <span>{params.row.edit?.extra?.isPublic ? 'Available' : 'None'}</span>
+        </Tooltip>
+    );
+};
 
 const microserviceStatusInfo = (params: GridRenderCellParams) => {
-    try {
-        const status = params.row.status[0].phase;
-        const checkStatus = status.toLowerCase();
+    let color = themeDark.palette.text.primary;
+    let icon = <QuestionMark sx={{ color }} />;
 
-        let color = themeDark.palette.text.primary;
-        let icon = <QuestionMark sx={{ color }} />;
+    const status = params.row.status[0]?.phase;
+    if (status === undefined) return 'N/A';
 
-        if (checkStatus.includes('running')) {
-            icon = <CheckCircleRounded />;
-        } else if (checkStatus.includes('pending')) {
-            color = themeDark.palette.warning.main;
-            icon = <WarningRounded sx={{ color }} />;
-        } else if (checkStatus.includes('failed')) {
-            color = themeDark.palette.error.main;
-            icon = <ErrorRounded sx={{ color }} />;
-        }
+    const checkStatus = status.toLowerCase();
 
-        return (
-            <Box sx={styles.status}>
-                {icon}
-                <Typography sx={{ ...styles.statusTitle, color }}>{status}</Typography>
-            </Box>
-        );
-    } catch (err) {
-        console.error(`Error with '${params.row.name}' status.`);
-        return '';
+    if (checkStatus.includes('running')) {
+        icon = <CheckCircleRounded />;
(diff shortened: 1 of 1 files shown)


Source/SelfService/Web/microservice/dataTable.tsx
```diff
@@ -28,25 +28,34 @@ const styles = {
...
+const tooltipCell = (params: GridRenderCellParams) => {
+    const publicUrl = params.row.edit?.extra?.isPublic;
+    if (publicUrl === undefined) return 'N/A';
+
+    return (
+        // TODO: Tooltip needs public urls. Map them into title.
+        <Tooltip title={``} arrow>
+            <span>{params.row.edit?.extra?.isPublic ? 'Available' : 'None'}</span>
+        </Tooltip>
+    );
```
jakhog, line 47: If we return just a string here, the styling might be different. I believe it is better to just change the text inside the tooltip and span:
```suggestion
    const hasPublicUrl = params.row.edit?.extra?.isPublic;

    return (
        // TODO: Tooltip needs public urls. Map them into title.
        <Tooltip title={``} arrow>
            <span>{
                hasPublicUrl === true ? 'Available' :
                hasPublicUrl === false ? 'None' :
                'N/A'
            }</span>
        </Tooltip>
    );
```

Source/SelfService/Web/microservice/dataTable.tsx
```diff
@@ -28,25 +28,34 @@ const styles = {
...
     return `${runtimeVersion || 'None'}`;
 };
 
-const tooltipCell = (params: GridRenderCellParams) => (
-    // TODO: Tooltip needs public urls. Map them into title.
-    <Tooltip title={``} arrow>
-        <span>{params.row.edit.extra.isPublic ? 'Available' : 'None'}</span>
-    </Tooltip>
-);
+const tooltipCell = (params: GridRenderCellParams) => {
```
jakhog, line 38: Lets rename this to:
```suggestion
const publicUrlCell = (params: GridRenderCellParams) => {
```

Source/SelfService/Web/microservice/dataTable.tsx
```diff
@@ -165,7 +175,7 @@ export const DataTable = ({ application, environment, microservices }: DataTable
             flex: 1,
             renderCell: statusCell,
             sortComparator: (v1, v2, param1, param2) =>
-                param1.value[0].phase.localeCompare(param2.value[0].phase)
+                param1.value[0]?.phase?.localeCompare(param2.value[0]?.phase)
```
jakhog, line 178: I believe this function might return `undefined` - which can cause undefined behaviour. So I suggest that we change this a little bit.

First, we can define a set of statuses that we care about - and then return them from the backend status:
```typescript
enum MicroserviceStatus {
    Running = 0,
    Pending = 1,
    Failing = 2,
    Unknown = 3,
};

const getMicroserviceState = (phase?: string): MicroserviceStatus => {
    const checkStatus = phase?.toLowerCase?.();

    if (typeof checkStatus !== 'string') {
        return MicroserviceStatus.Unknown;
    } else if (checkStatus.includes('running')) {
        return MicroserviceStatus.Running;
    } else if (checkStatus.includes('pending')) {
        return MicroserviceStatus.Pending;
    } else if (checkStatus.includes('failed')) {
        return MicroserviceStatus.Failing;
    }

    return MicroserviceStatus.Unknown;
}
```

Then we can change the `statusCell` to use that status for rendering:
```typescript

const statusCell = (params: GridRenderCellParams) => {
    let color = themeDark.palette.text.primary;
    let icon = <QuestionMark sx={{ color }} />;

    switch (getMicroserviceState(params.row.status[0]?.phase)) {
        case MicroserviceStatus.Running:
            icon = <CheckCircleRounded />;
            break;
        case MicroserviceStatus.Pending:
            color = themeDark.palette.warning.main;
            icon = <WarningRounded sx={{ color }} />;
            break;
        case MicroserviceStatus.Failing:
            color = themeDark.palette.error.main;
            icon = <ErrorRounded sx={{ color }} />;
            break;
        case MicroserviceStatus.Unknown:
            return 'N/A';
    }

    return (
        <Box sx={styles.status}>
            {icon}
            <Typography sx={{ ...styles.statusTitle, color }}>{status}</Typography>
        </Box>
    );
}
```

Then, lastly - we can change this sortComparator to sort using the numeric value of that enum we created:
```typescript
            sortComparator: (_, __, left, right) => {
                const leftStatus = getMicroserviceState(left.value[0]?.phase);
                const rightStatus = getMicroserviceState(right.value[0]?.phase);
                return leftStatus - rightStatus;
            }
```

This should result in a stable sorting. If you want to change the order of which things are sorted (maybe we want the errors on top?) - you can just rearrange the numbers in this enum:
```typescript
enum MicroserviceStatus {
    Running = 0,
    Pending = 1,
    Failing = 2,
    Unknown = 3,
};
```

jakhog (changes requested):
Some small suggestions - the big one relating to sorting on the status.

Source/SelfService/Web/microservice/dataTable.tsx
```diff
@@ -28,25 +28,34 @@ const styles = {
     }
 };
 
-export const sortByRuntimeVersion = (params: GridValueGetterParams) => {
-    const runtimeVersion = params.row.edit.extra.runtimeImage.replace(/dolittle\/runtime:/gi, '');
+const sortByRuntimeVersion = (params: GridValueGetterParams) => {
+    const runtimeVersion = params.row.edit?.extra?.runtimeImage?.replace(/dolittle\/runtime:/gi, '');
+    if (runtimeVersion === undefined) return 'N/A';
+
     return `${runtimeVersion || 'None'}`;
```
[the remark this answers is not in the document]
N00bG1rl: Well, it still needs to return 'None' if there is no runtime version. I could do something like that: runtimeVersion.replace(/^\w/, (c) => c.toUpperCase());

Would that be better solution?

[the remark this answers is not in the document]
jakhog: Hmm, I don't understand this. If there is no runtime version - wouldn't the variable `runtimeVersion` be `undefined`? Or am I missing something here..?

jakhog (approved):

PR #232, opened 2022-08-18
Change bg symbol color

N00bG1rl:
## Summary

Added background color for body what resolves white loading screen.

Changed background Dolittle logo color for having more contrast.

Screenshot 2022-08-18 at 11 36 06

### Added

- Body background color

### Changed

- Background logo color


jakhog (changes requested):
We can get the same effect by just changing the SVG manually. See: https://github.com/dolittle-platform/Authentication/pull/30#pullrequestreview-1077574131

Source/SelfService/Web/theme/assets/logos/symbol.svg
```diff
@@ -1,15 +1,11 @@
 <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 127.56 131.07">
     <defs>
         <linearGradient id="symbol-gradient" x1="0" y1="0" x2="0" y2="1">
-            <stop offset="0%" stop-color="#191A21" stop-opacity="1" />
-            <stop offset="100%" stop-color="#191A21" stop-opacity="0" />
+            <stop stop-color="#202128" stop-opacity="0.9"/>
+            <stop offset="1" stop-color="#23252E" stop-opacity="0.05"/>
```
jakhog, line 5: Reeally close now - but according to Figma, these stop opacities should be `1` and `0`
image

```suggestion
            <stop stop-color="#202128" stop-opacity="1"/>
            <stop offset="1" stop-color="#23252E" stop-opacity="0"/>
```

jakhog (changes requested):

Source/SelfService/Web/theme/assets/logos/symbol.svg
N00bG1rl: From where to you get this information? I am looking svg code and it shows opacity as 0.9 and 0.05.

Screenshot 2022-08-22 at 11 22 15

N00bG1rl: Okay, found it in Figma. :) There seems to be different between smaller part and bigger part. You think that we should add same gradient for both of them?

jakhog: In the SVG that is currently in the repo - the gradient applies to both of them. So there should only be a need for one :)

jakhog (approved):

PR #233, opened 2022-08-18
Capitalize first letter in runtime vs

N00bG1rl:
## Summary
Capitalized first letter in data table runtime version column.

Screenshot 2022-08-18 at 14 57 00

### Fixed

- Capitalized first letter in runtime version

### DIFF
--- a/Source/SelfService/Web/microservice/dataTable.tsx
+++ b/Source/SelfService/Web/microservice/dataTable.tsx
@@ -27,9 +27,9 @@ const styles = {
     }
 };
 
+const capitalize = (string: string) => string.charAt(0).toUpperCase() + string.slice(1);
 const sortByRuntimeVersion = (params: GridValueGetterParams) => {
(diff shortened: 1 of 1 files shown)


Source/SelfService/Web/microservice/dataTable.tsx
```diff
@@ -29,8 +29,9 @@ const styles = {
 
 const sortByRuntimeVersion = (params: GridValueGetterParams) => {
     const runtimeVersion = params.row.edit?.extra?.runtimeImage?.replace(/dolittle\/runtime:/gi, '');
+    const capitalize = (string: string) => string.charAt(0).toUpperCase() + string.slice(1);
```
jakhog, line 32: This function should be moved outside of the `sortByRuntimeVersion()` body - so that it is not recreated for each call. Probably a very small performance gain - but good practice :)

Also - this fix is possible to do in CSS by setting `text-transform: capitalize` on the component. This _could_ be considered a styling thing - but I'm not entirely sure which one I prefer. I'll let you pick.

jakhog (changes requested):

jakhog (approved):

PR #238, opened 2022-09-01
Remove insight menu item

N00bG1rl:
## Summary

Commented out all code related to 'Insight' page doe need for needing improvements (access the db directly instead of using APIs) etc.

### Removed

- Hided Insight menu item and its content.


jakhog (approved):
You could also just delete the code that is not used anymore (instead of comments) - but I don't see any issues here :)

PR #245, opened 2022-10-05
Fix Graph size calculation bug caused by scrollbar appearing.

jakhog:
## Summary

The `Vega` graph components use the `window:resize` event to re-calculate their own size. However, it turns out that this event **is not triggered** when a scrollbar is shown or hidden. This lead to some graphs having the wrong size if more content was loaded later and making the scrollbar appear.

This little fix listens to the `window.visualViewport:resize` event, which **is triggered** when a scrollbar is shown or hidden, and re-dispatches a new `window:resize` event. This forces `Vega` graphs to recalculate as normal, and it shouldn't cause issues for other elements.

### Fixed

- The `Vega` graphs did not re-calculate their size when a scrollbar appeared, which meant that for small windows the `CPU Usage` graph would be a little too wide when the `Memory Usage` graph appeared.

### DIFF
--- a/Source/SelfService/Web/App.tsx
+++ b/Source/SelfService/Web/App.tsx
@@ -8,6 +8,7 @@ import { LicenseInfo } from '@mui/x-license-pro';
 
 import { themeDark } from '@dolittle/design-system';
 
+import { useViewportResize } from './utils/useViewportResize';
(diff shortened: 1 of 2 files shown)


jakhog:
I should probably do something to not trigger the event twice when the `window` actually does resize...

PR #246, opened 2022-10-05
Fix bug where Environment selector on MicroserviceView screen shows a blank screen

jakhog:
## Summary

Fixes a bug where switching environments on the `MicroserviceView` (and other) screens causes a blank page to be a shown. Turns out it was not filtering microservices from the `MicroserviceStore` correctly - which caused it to partially load data for the wrong microservice. This is definitely not a good fix - but solves the problem for now.

### Changed

- The Mock-server now behaves more like the actual production APIs

### Fixed

- Getting Microservice-data from the store did not filter on environment, causing redirects to the overview not happening.


joelhoisko (approved):
:shipit:

PR #252, opened 2022-10-14
Bug-applications-list

woksin:
## Summary

When logging in to studio and there are no applications in the tenant you were logged in for you'd get a black screen. Now instead you will get an empty list and the choice to create a new application or go back to choose another tenant.

### Fixed

- Bug when no applications existed in the tenant

Before fix:
Screenshot 2022-10-14 at 21 01 37

After Fix:
Screenshot 2022-10-14 at 21 27 17


N00bG1rl (approved):
Great work!

PR #263, opened 2022-11-30
Update deployment to function for a monorepo

pavsaund:
## Summary

- Add the 'output' flag to output the contents to the 'standalone' folder - allowing the Docker build to expose this as a node.js server.
- Update the static output as well to reflect the output of the deployment. 
- Add the experimental feature 'outputFileTracingRoot' to be able to reference packages outside the directory root of the project.  Instead, it's pointing at the root of the monorepo.

https://nextjs.org/docs/advanced-features/output-file-tracing#caveats


pavsaund:
Looking better, at least - merging into main to see how the deployment works

PR #280, opened 2022-12-14
Env variables section

N00bG1rl:
## Summary

Added Configuration page Environment Variables section. User can add, delete, download and edit Environment Variables.

Screenshot 2022-12-14 at 07 29 27

Screenshot 2022-12-14 at 07 30 52


pavsaund (changes requested):
Functionality review (I haven't looked at the code yet):
- ~The values aren't stored in the env-configmap.yaml.~ It looks like the values are stored, but returned in the document coming back. (Same behaviour in prod). Perhaps worth looking at the desired behaviour here with Liz and creating a follow-up task
image
image

- Downloading secret env-variables fails. I don't know why this is happening, and I see the behaviour is weird in the currently deployed version. Perhaps we should register a separate bug for downloading of secret env vars?

- Editing a cell and making no changes sends a "save" command to the backend. This should not be needed.

- Adding an empty variable gives an error and removes the entry from the table. I think it's better to 1: not remove the edit field and 2: mark the cell in question as invalid with ex: a red-border. Perhaps something to discuss with Liz? Can also be solved as a separate story

Source/SelfService/Web/microservice/microserviceView/configuration/configuration.tsx
```diff
@@ -27,57 +19,28 @@ export type ConfigurationProps = {
...
-                <DownloadButtons
-                    environment={environment}
-                    microserviceName={currentMicroservice.name}
-                    applicationId={applicationId}
-                />
-            </Box>
-        </>
-    );
-};
+export const Configuration = ({ application, applicationId, environment, microserviceId, currentMicroservice }: ConfigurationProps) => (
```
pavsaund, line 22: The `Configuration`-component now looks like a good container component that deals with layout and structure of the page. On that note I suggest moving the `Accordion` components outside the different `*Section`-components. 

Doing so will also give the container page the ability to make decisions to example, only expand a single accordion at a time (if that is ever a need).

ex: 
```
<Accordion {...props }>
    <SetupSection  {...props } /> 
</Accordion>

<Accordion {...props }>
    <FilesSection  {...props } /> 
</Accordion>

//etc...
```

Source/SelfService/Web/microservice/microserviceView/configuration/configuration.tsx
```diff
@@ -27,57 +19,28 @@ export type ConfigurationProps = {
...
+    <>
+        <SetupSection
```
pavsaund, line 24: The component is called `SetupSection`, but the sub-folder and filname is `editSection`. I personally think `setup` is a better name than `edit`. Regardless, the filename and/or component name should be updated to match :)

Source/SelfService/Web/microservice/microserviceView/configuration/emptyDataTable.tsx
```diff
@@ -3,27 +3,28 @@
...
 import { AddCircle } from '@mui/icons-material';
 
 import { Button } from '@dolittle/design-system/atoms/Button';
 
-type NoConfigFilesProps = {
+type EmptyDataTableProps = {
+    title: string;
+    description: string;
+    buttonText: string;
+    sx?: SxProps
```
pavsaund, line 15: Does this component need to expose `sx`?

Source/SelfService/Web/microservice/microserviceView/configuration/emptyDataTable.tsx
```diff
@@ -3,27 +3,28 @@
...
     handleOnClick: () => void;
 };
 
-export const NoConfigFiles = ({ handleOnClick }: NoConfigFilesProps) => {
+export const EmptyDataTable = ({ title, description, buttonText, sx, handleOnClick }: EmptyDataTableProps) => {
```
pavsaund, line 19: I like this renaming. I think I see that this "empty-state" component can be a valuable component to use in a DataGrid molecule-component exposed by the design system. But I thin that may require some more thinking about

Source/SelfService/Web/microservice/microserviceView/configuration/filesSection/buttonGroup.tsx
```diff
@@ -34,7 +34,7 @@ export const ButtonGroup = ({ filePrompt, deleteDisabled, downloadDisabled, hand
 
         <Button
             variant='text'
-            label={`Download Configuration Maps`}
+            label={`Download files yaml`}
```
pavsaund, line 37: The button text here feels wrong. The yaml that is downloaded is a `ConfigMap, [as you can see here on dolittle.io](https://dolittle.io/docs/platform/update_configurations/). I know we probably want to move away from this, but perhaps a better wording would be: 
```suggestion
            label={`Download Configuration Files Yaml`}
```

Source/SelfService/Web/microservice/microserviceView/configuration/configuration.tsx
```diff
@@ -27,57 +19,28 @@ export type ConfigurationProps = {
...
+            application={application}
+            applicationId={applicationId}
+            environment={environment}
+            currentMicroservice={currentMicroservice}
+            microserviceId={microserviceId}
+        />
+
+        <FilesSection
```
pavsaund, line 32: For consistency, could the component and section be named `ConfigurationFilesSection`?

Source/SelfService/Web/microservice/microserviceView/configuration/environmentVariablesSection/environmentVariableSection.tsx
```diff
@@ -0,0 +1,288 @@
...
+            editable: true
+        },
+        {
+            field: 'isSecret',
+            headerName: 'Secret',
+            type: 'singleSelect',
+            valueOptions: [{ value: true, label: 'Yes' }, { value: false, label: 'No' }],
+            editable: true,
+            renderCell: ({ value }) => (
+                // TODO: Needs improvement
```
pavsaund, line 63: What kind of improvement is needed? Is it something to create a follow-up task for, or could we look at this together, perhaps?

Source/SelfService/Web/microservice/microserviceView/configuration/environmentVariablesSection/environmentVariableSection.tsx
```diff
@@ -0,0 +1,288 @@
...
+            [id]: { mode: GridRowModes.View, ignoreModifications: true },
+        });
+
+        const editedRow = envVariableTableRows.find(row => row.id === id);
+
+        if (editedRow!.isNew) {
+            setEnvVariableTableRows(envVariableTableRows.filter(row => row.id !== id));
+        }
+
+        // TODO: Check if there are any changes before updating
```
pavsaund, line 127: Aha, this looks like it's related to a previous comment I made earlier. So, don't attempt to save if the values haven't changed, right? Same comment as above, we can solve this in a follow-up PR or look at this together, perhaps?

Source/SelfService/Web/microservice/microserviceView/configuration/environmentVariablesSection/environmentVariableSection.tsx
```diff
@@ -0,0 +1,288 @@
...
+            enqueueSnackbar('File not deleted. Please try again.', { variant: 'error' });
+        }
+    };
+
+    // TODO: This is reused. consider moving
+    const configMapPrefix = `${environment.toLowerCase()}-${microserviceName.toLowerCase()}`;
+
+    const handleSecretEnvVariableDownload = () => {
+        const secretName = `${configMapPrefix}-secret-env-variables`;
+        const href = `${getServerUrlPrefix()}/live/application/${applicationId}/secret/${secretName}?download=1&fileType=yaml`;
```
pavsaund, line 205: I would consider moving this to the `api`-layer. Then you can also move the `configmapPrefix` there as a const / helper

Source/SelfService/Web/microservice/microserviceView/configuration/environmentVariablesSection/environmentVariableSection.tsx
```diff
@@ -0,0 +1,288 @@
...
+        enqueueSnackbar(`'${secretName}.yaml' has been downloaded.`);
+        window.open(href, '_blank');
+    };
+
+    const handleEnvVariableDownload = () => {
+        const configMapName = `${configMapPrefix}-env-variables`;
+        const href = `${getServerUrlPrefix()}/live/application/${applicationId}/configmap/${configMapName}?download=1&fileType=yaml`;
```
pavsaund, line 212: Same comment as above: I would consider moving this to the `api`-layer. Then you can also move the `configmapPrefix` there as a const / helper

Source/SelfService/Web/microservice/microserviceView/configuration/environmentVariablesSection/environmentVariableSection.tsx
```diff
@@ -0,0 +1,288 @@
...
+                </Box>
+
+                <RestartInfoBox name={microserviceName} isAlertBoxOpen={restartInfoBoxIsOpen} handleDismiss={() => setRestartInfoBoxIsOpen(false)} />
+
+                {noEnvVariables ?
+                    <EmptyDataTable
+                        title='No environment variables yet...'
+                        description={`To add your first environment variable, select 'add variable'. Provide a name, value and set its secrecy.`}
+                        buttonText='Add Variable'
+                        handleOnClick={handleEnvVariableAdd} // TODO: Sometimes throws error when clicked
```
pavsaund, line 266: This TODO seems interesting 😅 - any thoughts into why it throws error sometimes?

Source/SelfService/Web/microservice/microserviceView/configuration/environmentVariablesSection/environmentVariableSection.tsx
```diff
@@ -0,0 +1,288 @@
...
+                        sx={{ mb: 8 }}
+                    /> :
+                    <DataTablePro
```
pavsaund, line 269: I think this usage shows that we probably could just call the DataGridPro directly. What do you think?

pavsaund (approved):
Some suggestions and thoughts. The only real blocker here is the DataGridPro component in the design system, and the License being exposed directly.

As usual, I'm approving as not to block progress, but reach out and we can walk through things together.

Source/SelfService/Web/microservice/microserviceView/configuration/environmentVariablesSection/environmentVariableSection.tsx
N00bG1rl: Just styling.

1. In 'edit' mode, the select cell took too much space and overall row got too high.
2. After exiting from 'edit' mode, button from next line got focus.
3. Button had some ugly ripple effects and was wrong size.

Source/SelfService/Web/microservice/microserviceView/configuration/emptyDataTable.tsx
N00bG1rl: At the moment let it be as it is in use? Will move this to Data Grid component at some point.

Screenshot 2022-12-12 at 16 55 57

Source/SelfService/Web/microservice/microserviceView/configuration/environmentVariablesSection/environmentVariableSection.tsx
N00bG1rl: Uhh, dunno. Probably.

Source/SelfService/Web/App.tsx
```diff
@@ -41,18 +41,15 @@ import { ContainerRegistryScreen } from './screens/containerRegistryScreen';
 import { M3ConnectorScreen } from './screens/m3connectorScreen';
 import { LogsScreen } from './screens/logsScreen';
 
-LicenseInfo.setLicenseKey('4a83cf3032e8518adfbe8694d092262dTz00ODUxMyxFPTE2OTEyMjE0ODgxODUsUz1wcm8sTE09c3Vic2NyaXB0aW9uLEtWPTI=');
+LicenseInfo.setLicenseKey(process.env.REACT_APP_DATA_GRID_PRO_KEY!);
```
pavsaund, line 44: Rename suggestion of variable, since it's an MUI key - not a data grid pro key:

```suggestion
LicenseInfo.setLicenseKey(process.env.REACT_APP_MUI_LICENSE_KEY!);
```

pavsaund, line 44: The call to the license key should be done once per app:

Copied from earlier comment:

> According to the [Licensing page of MUI](https://mui.com/x/introduction/licensing/#where-to-install-the-key), the key should be exposed in the specific project using the DataGridPro component, and only once in the application, so it shouldn't be called whenever we call the wrapper component.
>
> I suggest deleting this from the component in the Design system, and instead use it in Studio directly, and in storybook so it's only called once per project.

pavsaund (approved):
A few small comments on the license key usage :)

PR #301, opened 2023-01-20
Upgrade react router

pavsaund:
## Summary

This PR upgrades the react router to the latest version, and does cleans up a few things along the way. There are a few motivations for the upgrade at this time.
 - Prepare for introducing several `modules` with their own navigation
 - Ensure that Studio doesn't lag behind on versions, making upgrading even harder in the future
 - Clean up some of the routes / Nested routes along the way

The guide at https://reactrouter.com/en/6.7.0/upgrading/v5 was the main reference for helping along the way.

What is **not done** in this PR:
 - Fixing click-jacking of navigation url's (no more onClick) for simple navigation
 - removing `selfservice` as base-url
 - Change ApplicationChanger behaviour to updated pattern. The reason for this is the store does not reset microservices across applications, and thus merges them across application environments. This needs to be fixed as a separate PR.

### Added

- For local dev, added handling in webpack devserver for known urls not using /selfservice basename
- Added a `ButtonLink` component in Studio that extends the design system button with react-router`Link`-component

### Changed

- Upgraded react router to v6.6.2
- Now leaning on relative url's. from within a Nested route, simplifying navigation and the need to build up full urls for router-navigation
- `Button` has a new optional `overrides` property to allow overriding from external usage. Used by `ButtonLink`

### Fixed

- microserviceTable now re-renders when microservices change, and sets the isLoading flag before and after.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1203789530758097

### DIFF
--- a/Source/DesignSystem/atoms/Button/Button.stories.tsx
+++ b/Source/DesignSystem/atoms/Button/Button.stories.tsx
@@ -18,17 +18,17 @@ export default {
             description: {
                 component: `A button triggers an event or action. Its label should let the user know what will happen next.
 
-**Styling:** Buttons come in three style variations. All variations can be used with or without icons to help lift the UI and 
+**Styling:** Buttons come in three style variations. All variations can be used with or without icons to help lift the UI and
 quickly visually communicate to the user what the button will do.
 
-- *Filled buttons* are reserved for primary actions. They should be used to call attention to an action on a form or to 
-highlight the strongest call to action on the page. Filled buttons should appear only once per container (not including the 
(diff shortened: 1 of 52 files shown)


N00bG1rl (approved):
Looks good, well done!

Local environment '.auth' routes does not work at the moment.

pavsaund:
I've done a pretty thorough test of this in dev and ironed out the bugs that I suggest merging it in and dealing with any remaining regressions as they turn up

PR #306, opened 2023-02-02
change runtime version 8 to 8.8.2

TomasEkeli:
this puts new services on the at-this-time latest runtime version

## Summary

change the "8" version of the runtime to 8.8.2, which is currently the latest

not 100% sure that nothing else needs to change - so a review would be nice :)


pavsaund (approved):
Perfect, thanks!

PR #313, opened 2023-02-16
Fix broken links to images tags in container registry view for images containing '/'

pavsaund:
## Summary

As part of the upgrade to latest react router (#301) the image/tags route was changed and no longer supported using `/` as part of the URL. This fix encodes the :image parameter so that it should work

### Fixed

- Fix broken links to images tags in container registry view for images containing '/'


pavsaund:
Confirmed this works in dev.
image

PR #324, opened 2023-03-08
Introduce SET_HEADERS env var for configuring headers to be set

eliasgs:
## Summary

Introducing the `SET_HEADERS` environment variable for configuring custom headers for routes. The value of this is expected to be a valid json document, where the keys are the `http.ServeMux` pattern to set headers for, and the value are a json document mapping from http header to [text/template](https://pkg.go.dev/text/template) value (this template is then executed with the incoming request as the data argument). Note: The pattern MUST exactly match a route defined in the `PROXY` environment variable. Also if no `PROXY` is defined, `SET_HEADERS` is ignored.

Example:
```
SET_HEADERS='{"/bridge/": {"X-organisation-id": "{{ .Header.Get \"Tenant-ID\" }}", "Tenant-ID": "..."}}'
```

### Added

- `SET_HEADERS` environment variable
- `parseSetHeaders` for parsing the value of `SET_HEADERS`
- functionality for setting headers on the downstream request in the `Proxy` function

### Changed

- `Proxy` now needs both the `pattern` and the `host` as arguments

### DIFF
--- a/Source/SelfService/Backend/main.go
+++ b/Source/SelfService/Backend/main.go
@@ -1,18 +1,23 @@
 package main
 
 import (
+	"bytes"
 	"encoding/json"
 	"log"
 	"net/http"
 	"net/http/httputil"
 	"net/url"
 	"os"
+	"text/template"
 	"time"
 
 	"github.com/sirupsen/logrus"
 )
 
 type Routes map[string]string
+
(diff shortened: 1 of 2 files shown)


TomasEkeli (approved):
nice tests :)

PR #424, opened 2023-07-14
Add streams to expose data

TomasEkeli:
## Summary

Adds a new sub-page for consuming event streams that link to the async-api -specification for the connection.

### additional

- sets up k3d in the devcontainer and adds a new version of the reset-local-k3d -script (which works with recent versions of k3d)
- re-generated code from the bridge-api open-api -spec (swagger) to one that includes the KafkaServiceAccount -models


TomasEkeli:
as a general comment: it is hard to get the studio working locally connected to the bridge-api locally. we should attend to this.

pavsaund:
> as a general comment: it is hard to get the studio working locally connected to the bridge-api locally. we should attend to this.

I tested this out and it seems to be related to both Studio and Bridge being run in their respective devcontainers. It seems like they can't communicate with each other. When I fire up Studio outside the devcontainer it works just fine.

pavsaund:
Updated the branch to use the `api/bridge` path, as this use the backend to proxy the request. Also made the URL based on current location, and moved the logic to its own tab

PR #434, opened 2023-08-02
Add env selectors

N00bG1rl:
## Summary

Add environment selections and display environment column in MicroservicesDataGrid.

### Added

- Environment selection dropdown to DeployMicroservice
- Environment selection dropdown to Microservice Configuration page
- DataGrid default props to Design System
- DataGridWrapper component to Design System
- Environment column to MicroservicesDataGrid
- (New)Switch component to Design System

### Fixed

- Logs page layout on smaller screens
- Logs line elements alignment

### Removed

- Redundant code


Source/SelfService/Web/applications/microservice/microservices/microservicesDataGridColumns.tsx
```diff
@@ -42,13 +42,23 @@ const StatusCell = (params: GridRenderCellParams<any, HealthStatusTableRowProps[
...
         flex: 1,
     },
+    {
+        field: 'environment',
+        headerName: 'Environment',
+        minWidth: 170,
+        flex: 1,
+        valueGetter: ({ row }: HealthStatusTableRowProps) => {
+            const environment = row.environment;
+            return environment === 'Prod' ? 'Production' : environment === 'Dev' ? 'Development' : environment;
```
pavsaund, line 59: We cannot do this, as we do not decide which environment names a user has. Also, it is quite common to have a Test environment as well. For example some customers have Test and Demo, as well as Dev & Prod

pavsaund: To be clear, it should be limited on the input, not the output. IE: When the user creates an application. Developers should recognise the env names as it is in the hosted cluster.

Source/SelfService/Web/applications/microservice/microservices/microservicesDataGrid.tsx
```diff
@@ -2,35 +2,37 @@
...
     environment: string;
     application: HttpResponseApplication;
     microservices: MicroserviceObject[];
 };
 
-export const MicroserviceTable = ({ application, environment, microservices }: MicroserviceTableProps) => {
+export const MicroservicesDataGrid = ({ application, environment, microservices }: MicroservicesDataGridProps) => {
     const navigate = useNavigate();
 
     const [microserviceRows, setMicroserviceRows] = useState<(MicroserviceObject | undefined)[]>([]);
```
pavsaund, line 26: is this the correct type? can the array containe undefined and MicroserviceObject entries, or is it meant to be `MicroserviceObject[] | undefined`?

pavsaund (approved):
Looks like some nice improvements. I've left a few comments. Let me know what you think?

Source/SelfService/Web/applications/microservice/microservices/microservicesDataGrid.tsx
N00bG1rl: I really do not know. It would take too much time to delve into it. I don't think it's worth it... I know it gives errors that way.

PR #490, opened 2023-10-24
Switched service listings to use new endpoint with pod status included

mhelleborg:
## Summary

Changed microservices to use the new endpoints which include pod status when listing services. This means each service does not need to individually call getpodstatus, and can render faster.

The new endpoint also uses the images which are actually used by the running pod, which means there will no longer be a stale runtime version.

### DIFF
--- a/Source/SelfService/Web/applications/microservice/microservicesOverview/microservicesList/microservicesDataGrid.tsx
+++ b/Source/SelfService/Web/applications/microservice/microservicesOverview/microservicesList/microservicesDataGrid.tsx
@@ -15,10 +15,11 @@ import { HttpResponseApplication } from '../../../../apis/solutions/application'
 import { microservicesDataGridColumns } from './microservicesDataGridColumns';
 
 import { getMicroserviceInfo } from '../../utils/getMicroserviceInfo';
+import { MicroserviceStore } from '../../../stores/microservice';
 
 export type MicroservicesDataGridProps = {
     application: HttpResponseApplication;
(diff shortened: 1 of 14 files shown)


Source/SelfService/Web/applications/microservice/microservicesOverview/microservicesList/microservicesDataGrid.tsx
```diff
@@ -32,12 +33,12 @@ export const MicroservicesDataGrid = ({ application, microservices }: Microservi
...
         Promise.all(microservices.map(async microservice => {
             const microserviceInfo = getMicroserviceInfo(application, microservice);
-            const status = await getMicroserviceStatus(microservice.id, microservice.environment);
+            // const status = await getMicroserviceStatus(microservice.id, microservice.environment);
 
             return {
                 ...microservice,
                 edit: microserviceInfo,
-                phase: status[0]?.phase,
+                phase: microservice.live.phase,
```
N00bG1rl, line 41: I believe that we don't need 'phase' here anymore :)

N00bG1rl (approved):

Source/SelfService/Web/applications/microservice/microservicesOverview/microservicesList/microservicesDataGrid.tsx
mhelleborg: We use it for the list view, so I think it's still needed :)

